PR #157, opened 2014-12-18
Remove ITypeActivator

halter73:
Add the methods from TypeActivator to the static ActivatorUtilities class
- Add ITypeActivator extension methods to ActivatorUtilities  as well
- This is preparation for removing TypeActivator and ITypeActivator


yishaigalatzer:
@muratg MVC has removed dependency on `ITypeActivator` we are no longer blocking this issue.

muratg:
This was done with a separate PR: https://github.com/aspnet/DependencyInjection/issues/190 

Stephen, it should be OK to close this one.

Issue #158, opened 2014-12-31
Add component that deals with the abstraction between callcontext/asynclocal


avanderhoorn:
As a developer, in cases where I need to access request scoped data from within singleton instances, I would like a component that deals with the abstraction between callcontext/asynclocal - `IScopedData<T>`. I could do this manually myself, but this seems like something that should be commonly handled.

This came out of a discussion I had with @davidfowl on Jabbr.

Eilon:
Moving this to Backlog as we are not planning to make any changes for this at this time.

davidfowl:
Doesn't need to be built in.

PR #159, opened 2015-01-09
Rename IContextAccessor -> IScopeInstance

HaoK:
cc @yishaigalatzer @divega @davidfowl @lodejard


src/Microsoft.Framework.DependencyInjection/project.json
```diff
@@ -15,7 +15,7 @@
             "dependencies": {
                 "System.Collections.Concurrent": "4.0.10-beta-*",
                 "System.Linq.Expressions": "4.0-beta-*",
-                "System.Threading": "4.0.10-beta-*",
+                "System.Threading": "4.0.0-beta-*",
```
davidfowl: ?

davidfowl:
The only push back I have on this is https://github.com/aspnet/DependencyInjection/issues/158. We should discuss the naming of this new thing in the same context.

src/Microsoft.Framework.DependencyInjection/project.json
HaoK: Oops bad merge I will fix

HaoK:
Does supporting #158 require any changes to the interface?  Or do we just need to make sure that the name convey's that it can also be used for that scenario?

davidfowl:
@HaoK no changes are required to the interface itself. It might even be another interface name. My only concern was calling both things IScopedSomething<T> might be confusing.

HaoK:
Got it, yeah having IScopeLocal and IScopeData which do different things would be pretty confusing.

Could we rename this one to be IRequestLocal then?  The scope is implicitly the request today isn't it?

davidfowl:
Maybe we just rename the other one?

HaoK:
Yeah the other one isn't specifically related to scope, that one could be ICallContext maybe?  Since its abstracting async local/call context?

davidfowl:
Hmm

halter73:
Would `IScopeValue` be a better name? That way it might be less likely to get confused with whatever we end up doing to abstract the call context.

davidfowl:
:shipit:

davidfowl:
Value seems better tbh

HaoK:
Maybe IScopeInstance.Value?   IScopeValue is ok too

davidfowl:
```C#
IScopedInstance<T> 
```

Sounds better

davidfowl:
:shipit:

yishaigalatzer:
IScopeValue sounds as good as it is going to get :)

HaoK:
I think we settled on IScopedInstance

On Jan 10, 2015, at 4:29 PM, "Yishai Galatzer" <notifications@github.com<mailto:notifications@github.com>> wrote:


IScopeValue sounds as good as it is going to get :)

-

yishaigalatzer:
That's fine too

HaoK:
Merged b3f137eaab3366f3e03b047d1ff6fb40bb26cbdc

PR #160, opened 2015-01-15
Castle.Windsor adapter without scoping support.

### LINKED ISSUE #24: Add utility method to Populate Castle Windsor
- this populates and duck-punches the adjustments for required behaviors. 
- `void static CastleWindsorFooStatic.Configure(IEnumerable<IServiceDescriptor> services, IWindsorContainer container)`

egorpavlikhin:
Scope in Castle.Windsor only defines when an instance will be released, and nothing else, so it's not the same concept as in Ninject and Autofac. There isn't an easy way to implement similar functionality on top of WindsorContainer. Fixes #24

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection.Windsor/WindsorRegistration.cs
+++ b/src/Microsoft.Framework.DependencyInjection.Windsor/WindsorRegistration.cs
@@ -1,0 +1,159 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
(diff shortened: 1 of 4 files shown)


halter73:
It looks like Castle.Windsor uses the CallContext to maintain scope: http://docs.castleproject.org/Windsor.LifeStyles.ashx#Standard_lifestyles:_the_less_common_ones_15

@lodejard @davidfowl Do you think that implementing scoping this way is sufficient?

moreira-daniel:
Is there a special and important reason for choosing the Autofac or Castle? Considered other like Simple Injector?

davidfowl:
We've moved all 3rd party containers out of this repository. It would be great to contribute this to the Castle itself

PR #165, opened 2015-01-30
Modify ActivatorUtilities.CreateFactory to return a delegate instead of a Func

### LINKED ISSUE #164: Provide a nicer delegate for `ActivatorUtilities.CreateFactory`
It'd be much nicer to have a delegate returned as a result of `ActivatorUtilties.CreateFactory` rather than returning a raw Func as it does today:

```C#
public delegate object ObjectFactory(IServiceProvider, object[]);

```

The resulting code at the consumer end looks a lot nicer:

```C#
private IDictionary<Type, Func<IServiceProvider, object[], object>> _objectFactories = 
     new Dictionary<Type, Func<IServiceProvider, object[], object>>();
```
versus

```C#
private IDictionary<Type,ObjectFactory> _objectFactories = new Dictionary<Type, ObjectFactory>();
```

pranavkm:
Fixes #164

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection/ActivatorUtilities.cs
+++ b/src/Microsoft.Framework.DependencyInjection/ActivatorUtilities.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Framework.DependencyInjection
         }
 
         /// <summary>
-        /// Create a function that will instantiate a type with constructor arguments provided directly
+        /// Create a delegate that will instantiate a type with constructor arguments provided directly
         /// and/or from an <see cref="IServiceProvider"/>.
         /// </summary>
         /// <param name="instanceType">The type to activate</param>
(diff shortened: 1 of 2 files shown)


davidfowl:
Why?

rynowak:
Fixes ugliness when we create a `ConcurrentDictionary<Type, Func<IServiceProvider, object[], object>>

Aren't we avoiding putting `Func<...>` in signatures in general?

pranavkm:
The https://github.com/aspnet/DependencyInjection/issues/164 has details about this. It makes the consuming code look a lot nicer:

```C#
private IDictionary<Type, Func<IServiceProvider, object[], object>> _objectFactories = 
     new Dictionary<Type, Func<IServiceProvider, object[], object>>();
```
versus

```C#
private IDictionary<Type,ObjectFactory> _objectFactories = new Dictionary<Type, ObjectFactory>();
```

pranavkm:
Ideally, it'd be nice to have an overload that removes the `object[]` argument (so you don't have to do `Type.EmptyTypes` on the other end). But that's a bit more code change and we could live without that.

halter73:
:shipit:

PR #168, opened 2015-02-02
Added fix for #Issue167

### LINKED ISSUE #167: ActivatorUtilities CreateFactory does not throw an exception if constructor parameter is not registered
I have a class with a constructor that takes a service as a parameter and the service is not registered
```
    public class CreationCountFakeService
    { 
         public CreationCountFakeService(IFakeService dependency)
         {
           ....
         }
    }
```

For the above scenario, if I call `ActivatorUtilities.CreateFactory(...)`, `dependency` comes back as null instead of throwing an exception. `ActivatorUtilities.CreateInstance` throws InvalidOperationException

kirthik:
Fixed #167 
Added unit test for the fix
Split existing test into two and added right parameters in both cases (This test was failing with the fix. Hence I had to split it)

@halter73, @pranavkm - Could you please take a look?

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection/ActivatorUtilities.cs
+++ b/src/Microsoft.Framework.DependencyInjection/ActivatorUtilities.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Framework.DependencyInjection
     public static class ActivatorUtilities
     {
         private static readonly MethodInfo GetServiceInfo =
-            GetMethodInfo<Func<IServiceProvider, Type, bool, object>>((sp, t, c) => GetService(sp, t, c));
+            GetMethodInfo<Func<IServiceProvider, Type, Type, bool, object>>((sp, t, r, c) => GetService(sp, t, r, c));
 
         /// <summary>
         /// Instantiate a type with constructor arguments provided directly and/or from an <see cref="IServiceProvider"/>.
@@ -125,13 +125,15 @@ namespace Microsoft.Framework.DependencyInjection
             return mc.Method;
         }
 
-        private static object GetService(IServiceProvider sp, Type type, bool condition)
+        private static object GetService(IServiceProvider sp, Type type, Type requiredBy, bool isDefaultParameterRequired)
         {
-            if (condition)
+            var service = sp.GetService(type);
+            if (service == null && !isDefaultParameterRequired)
             {
-                return sp.GetService(type);
+                throw new InvalidOperationException(
+                Resources.FormatCannotResolveService(type, requiredBy));
             }
-            return sp.GetRequiredService(type);
+            return service;
         }
 
         private static Expression BuildFactoryExpression(
@@ -156,6 +158,7 @@ namespace Microsoft.Framework.DependencyInjection
                     var constructorParameterHasDefault = constructorParameters[i].HasDefaultValue;
                     var parameterTypeExpression = new Expression[] { serviceProvider,
                         Expression.Constant(parameterType),
+                        Expression.Constant(constructor.DeclaringType),
                         Expression.Constant(constructorParameterHasDefault) };
                     constructorArguments[i] = Expression.Call(GetServiceInfo, parameterTypeExpression);
                 }
@@ -309,8 +312,8 @@ namespace Microsoft.Framework.DependencyInjection
                             if (!_parameters[index].HasDefaultValue)
                             {
                                 throw new InvalidOperationException(Resources.FormatCannotResolveService(
-                                    _constructor.DeclaringType,
-                                    _parameters[index].ParameterType));
+                                    _parameters[index].ParameterType,
+                                    _constructor.DeclaringType));
                             }
                             else
                             {
--- a/src/Microsoft.Framework.DependencyInjection/ActivatorUtilities.cs
+++ b/src/Microsoft.Framework.DependencyInjection/ActivatorUtilities.cs
@@ -14,7 +14,8 @@ namespace Microsoft.Framework.DependencyInjection
     /// </summary>
     public static class ActivatorUtilities
     {
-        private static readonly MethodInfo GetServiceInfo = GetMethodInfo<Func<IServiceProvider, Type, object>>((sp, t) => sp.GetService(t));
+        private static readonly MethodInfo GetServiceInfo =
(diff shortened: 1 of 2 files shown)


msotclas:
Hi __@kirthik__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>
        In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla2.msopentech.com.
    </span>

TTYL, MSOTBOT;

src/Microsoft.Framework.DependencyInjection/ActivatorUtilities.cs
```diff
@@ -143,8 +144,8 @@ private static MethodInfo GetMethodInfo<T>(Expression<T> expr)
                 }
                 else
                 {
-                    var parameterTypeExpression = Expression.Constant(parameterType);
-                    constructorArguments[i] = Expression.Call(serviceProvider, GetServiceInfo, parameterTypeExpression);
+                    var parameterTypeExpression = new Expression[] { serviceProvider, Expression.Constant(parameterType) };
```
pranavkm: What changed here?

kirthik: Since GetRequiredService is static method as opposed to GetService, I could not use the overload that was being used. I had to change the overload

test/Microsoft.Framework.DependencyInjection.Tests/ActivatorUtilitiesTests.cs
```diff
@@ -79,12 +79,24 @@ public void TypeActivatorWorksWithStaticCtor(CreateInstanceFunc createFunc)
             Assert.NotNull(anotherClass);
         }
 
-        [Theory]
-        [MemberData(nameof(CreateInstanceFuncs))]
-        public void TypeActivatorWorksWithCtorWithOptionalArgs(CreateInstanceFunc createFunc)
```
halter73: This test shouldn't have broke.

`CreateInstance` will allow for missing services if there is a default value for the parameter. `CreateFactory` used to work the same way in this regard. Now it doesn't.

kirthik: Yes since I changed GetService to GetRequiredService, if type is not provided, it throws an exception saying type is not registered. Earlier, it would just return null.

test/Microsoft.Framework.DependencyInjection.Tests/ActivatorUtilitiesTests.cs
```diff
@@ -220,5 +232,16 @@ public void GetServiceOrCreateInstanceUnregisteredService()
             Assert.Equal(2, service.InstanceId);
             Assert.Equal(2, CreationCountFakeService.InstanceCount);
         }
+
+        [Fact]
+        public void CreateFactoryUnRegisteredServiceAsConstructorParameter()
```
halter73: Can we have this same test for `CreateInstance`? I think making this a theory would be better.

test/Microsoft.Framework.DependencyInjection.Tests/ActivatorUtilitiesTests.cs
halter73: This is not the way CreateInstance behaves.

You could fix the behavior by calling GetService instead of GetDefaultService iff `constructorParameters[i].HasDefaultValue` is true.

kirthik: True, CreateInstance does not but CreateFactory behaves differently which is why i had to split the test into 2 tests. CreateInstance test has exact same code as old test. For CreateFactory test, I added extra parameter( typeof(string))

test/Microsoft.Framework.DependencyInjection.Tests/ActivatorUtilitiesTests.cs
kirthik: Sure. Will add.

test/Microsoft.Framework.DependencyInjection.Tests/ActivatorUtilitiesTests.cs
halter73: I don't think you should have changed/split the test. It wouldn't be difficult to make the old test pass.

When you are fixing CreateFactory to match CreateInstance's behavior in one scenario, I don't think you should be breaking the consistency of another behavior unless there is a really good reason for doing so.

kirthik:
@halter73 - I updated the code so previous behavior of default values is retained. Please take a look

test/Microsoft.Framework.DependencyInjection.Tests/ActivatorUtilitiesTests.cs
```diff
@@ -220,5 +220,17 @@ public void GetServiceOrCreateInstanceUnregisteredService()
...
+
+        [Theory]
+        [MemberData(nameof(CreateInstanceFuncs))]
+        public void UnRegisteredServiceAsConstructorParameterThrowsException(CreateInstanceFunc createFunc)
+        {
+            var serviceProvider = new ServiceCollection()
+                .AddSingleton<CreationCountFakeService>()
+                .BuildServiceProvider();
+
+            Assert.Throws<InvalidOperationException>(() =>
```
halter73: We should assert that we get the expected exception message like we do in the other tests that use Assert.Throws.

Right now, the exception thrown CreateInstance has a different message than the one thrown from CreateFactory. The exception message from CreateInstance is a little better since it tells what type is being activated.

This will mean that you can no longer use GetRequiredService:

```C#
    private static object GetService(IServiceProvider sp, Type type, Type requiredBy, bool serviceRequired)
    {
        var service = sp.GetService(type);

        if (service == null && serviceRequired)
        {
            throw new InvalidOperationException(
                Resources.FormatCannotResolveService(type, requiredBy));
        }

        return service;
    }
```

I noticed while writing this comment that CreateInstance doesn't provide the format parameters for FormatCannotResolveService in the correct order, so you could fix that up too as part of this change.

halter73:
:shipit: after fixing the exception message and test.

kirthik:
Checked in - a534cd1de2fc813e04ddc14420f904ff53891f64

PR #172, opened 2015-02-07
Add IList<IServiceDescriptor> property to IServiceCollection

pranavkm:


pranavkm:
Based on a discussion with @loudej - the intent is to expose the backing list to allow removing service descriptors that were previously added.

davidfowl:
Why not just make IServiceCollection be an IList then?

pranavkm:
You end up having to implement quite a few more properties to make the type be an IList. This is less hassle to create.

pranavkm:
Updated.

src/Microsoft.Framework.DependencyInjection/ServiceCollection.cs
```diff
@@ -14,6 +14,12 @@ public class ServiceCollection : IServiceCollection
         private readonly List<IServiceDescriptor> _descriptors = new List<IServiceDescriptor>();
 
         /// <inheritdoc />
+        public int Count => _descriptors.Count;
```
davidfowl: Nice

davidfowl:
I know @halter73 spoke about adding a .Replace method as well.

pranavkm:
I'm actually using `Remove` to implement `Replace` in Mvc - I'll just move the code in here. The only issue is the `Replace` semantics isn't super clear - if you have more than service of the same type, is the Replace supposed to Replace all of them (i.e. RemoveAll + Insert) or just Replace the first one it encounters?

src/Microsoft.Framework.DependencyInjection/ServiceCollection.cs
lodejard: Wow

PR #175, opened 2015-02-11
Remove portable target

bricelam:
Given the removal of ANIs, this isn't needed anymore


pranavkm:
:shipit:

yishaigalatzer:
@bricelam is this merged?

davidfowl:
Why do this?

bricelam:
@yishaigalatzer Not merged yet, we still depend on it.
@davidfowl After aspnet/EntityFramework#1682 is merged, they won't be needed anymore. We may bring them back if we decide to target Windows Store/Phone 8.1, but I'm hoping for just Windows 10.

davidfowl:
@bricelam kewl. Can we nuke it now then?

PR #176, opened 2015-02-12
Remove [AssemblyNeutral] attributes

bricelam:


src/Microsoft.Framework.DependencyInjection/ServiceLookup/IServiceManifest.cs
```diff
@@ -6,9 +6,6 @@
 
 namespace Microsoft.Framework.DependencyInjection.ServiceLookup
```
pranavkm: This type needs to be moved to the Interfaces assembly.

davidfowl: Into the runtime.

bricelam: ...so to the `Microsoft.Framework.Runtime.Interfaces` package in the `Microsoft.Framework.DependencyInjection.ServiceLookup` namespace?

davidfowl: We change the namespace.

pranavkm: This is moving to KRuntime. cc @anurse

bricelam: Removed in 663f81d358ba9b3e7d2b6d69f409b7b2137a3350

PR #177, opened 2015-02-13
Make scoped instance IDisposable

yishaigalatzer:
Made the change to dispose content of `IScopedInstance` if the DI system disposes it.

Also some other minor cleanup on files I touched during this work.

CC / @divega @HaoK

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection/ServiceDescriptor.cs
+++ b/src/Microsoft.Framework.DependencyInjection/ServiceDescriptor.cs
@@ -57,18 +57,18 @@ namespace Microsoft.Framework.DependencyInjection
         }
 
         /// <inheritdoc />
-        public LifecycleKind Lifecycle { get; private set; }
+        public LifecycleKind Lifecycle { get; }
 
         /// <inheritdoc />
-        public Type ServiceType { get; private set; }
+        public Type ServiceType { get; }
 
         /// <inheritdoc />
-        public Type ImplementationType { get; private set; }
+        public Type ImplementationType { get; }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Framework.DependencyInjection/ServiceDescriptor.cs
```diff
@@ -57,18 +57,18 @@ private ServiceDescriptor(Type serviceType, LifecycleKind lifecycle)
         }
 
         /// <inheritdoc />
-        public LifecycleKind Lifecycle { get; private set; }
+        public LifecycleKind Lifecycle { get; }
```
yishaigalatzer: C#6 cleanup

test/Microsoft.Framework.DependencyInjection.Tests/ScopingContainerTestBase.cs
```diff
@@ -4,9 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
```
yishaigalatzer: ordering

HaoK:
:shipit: looks good, do we want to consider adding a psuedo functional test that verifies that all the scoped instances actually get disposed when the service provider goes away?

yishaigalatzer:
I was considering it. But didn't see functional tests for di, and didn't think this small change justifies it. Up to you

yishaigalatzer:
Merged 8d29c5b8d6539249ca4c97da8c63083f6b53a9af

PR #180, opened 2015-02-23
Fix order of parameters in CannotResolveService message

Daniel15:
`CannotResolveService` takes the service that can't be resolved as the first parameter, and the type requiring it as the second parameter. This is correct in `Service.CreateCallSite` and `ActivatorUtilities.GetService`, but the parameters were in the wrong order in `TypeActivator`.

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection/TypeActivator.cs
+++ b/src/Microsoft.Framework.DependencyInjection/TypeActivator.cs
@@ -240,8 +240,8 @@ namespace Microsoft.Framework.DependencyInjection
                             if (!_parameters[index].HasDefaultValue)
                             {
                                 throw new InvalidOperationException(Resources.FormatCannotResolveService(
-                                    _constructor.DeclaringType, 
-                                    _parameters[index].ParameterType));
+                                    _parameters[index].ParameterType,
+                                    _constructor.DeclaringType));
                             }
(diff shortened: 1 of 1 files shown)


msotclas:
Hi __@Daniel15__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>
        This seems like a small (but important) contribution, so no Contribution License Agreement is required at this point. Real humans will now evaluate your PR.
    </span>

TTYL, MSOTBOT;

davidfowl:
No test?

Daniel15:
This is a trivial change and I didn't see any existing tests for `TypeActivator`

Eilon:
@davidfowl while a unit test would sure be nice, it doesn't seem super necessary given the circumstances, no?

davidfowl:
I guess...

yishaigalatzer:
I thought we are removing this type altogether

yishaigalatzer:
#157

davidfowl:
We are

yishaigalatzer:
Thanks for the PR, as you can see we are going to delete this type (so don't depend on it ;))

So I'm going to close it, thanks again for caring.

PR #195, opened 2015-03-05
Migrate ServiceCollectionExtensions to Interfaces package

GrabYourPitchforks:
BuildServiceProvider remains since it relies on concrete ServiceProvider type
AddTypeActivator also remains, but it's disappearing soon

/cc @HaoK @davidfowl @Eilon

PR #196, opened 2015-03-05
Add more methods to the .Interfaces assembly

davidfowl:
- Moved ServiceCollectionExtensions
- Moved ServiceProviderExtensions
- Moved ITypeActivator (but it'll be gone soon)

/cc @HaoK


HaoK:
Ok, looks like this is the same as what levi's PR was :)  https://github.com/aspnet/DependencyInjection/pull/195/files

PR #211, opened 2015-03-27
Moved ServiceProviderExtensions from namespace Microsoft.Framework.DependencyInjection to `System.

david-driscoll:
This is so that `Microsoft.Framework.DependencyInjection` doesn't have to be added as a using every time you to get a strongly typed value from IServiceProvider.

It's entirely possible there is a very good reason that this isn't in the `System` namespace already, but this is one of the small pain points I currently encounter when working with Aspnet 5 day to day.

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection.Interfaces/ServiceProviderExtensions.cs
+++ b/src/Microsoft.Framework.DependencyInjection.Interfaces/ServiceProviderExtensions.cs
@@ -6,8 +6,9 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Framework.DependencyInjection.Interfaces;
 using Microsoft.Framework.Internal;
+using Microsoft.Framework.DependencyInjection;
 
-namespace Microsoft.Framework.DependencyInjection
+namespace System
 {
     public static class ServiceProviderExtensions
     {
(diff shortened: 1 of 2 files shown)


msotclas:
Hi __@david-driscoll__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>
        In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla2.msopentech.com.
    </span>

TTYL, MSOTBOT;

msotclas:
__@david-driscoll__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, MSOTBOT;

Eilon:
@KrzysztofCwalina do you have thoughts on this?

KrzysztofCwalina:
Is there some namespace that is often already imported in ASP projects? 

The problem with adding these APIs to System root is that now the type would be more visible to many developers that might now be using it. For example, MSDN System namespace list would probably have to now include it.

Also, the same argument for adding it to System could be used for any type, and yet we don't have all types in one giant flat namespace.

Eilon:
@KrzysztofCwalina I think in this particular case that DI namespace is the one that's already imported. I think these APIs are used most frequently in two scenarios:

1. The app's Startup class, which already imports this namespace (that's where all the other "AddXyz()" methods are for registering MVC, EF, Identity, etc. services)
2. People writing libraries that extend these similar APIs

I'm not sure I see the need of putting this in `System`.

jayharris:
Though I appreciate the pain and the idea, I'm not a fan of extension methods on `System`. If the method was `System` important, it should be on the core Interface. Since it only warrants an extension method, as a developer I want to be able to not use it--to turn it off--and to use my own implementation instead. This becomes much more difficult when my extension method is automatically in conflict with the available-by-default `System` extension.

Eilon:
Based on current feedback, closing this.

PR #226, opened 2015-05-05
fixed structuremap implementation

MrKevHunter:
I've added StructureMap 3 support, let me know if there is anything wrong.

### DIFF
--- a/DependencyInjection.sln
+++ b/DependencyInjection.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.22803.0
+VisualStudioVersion = 14.0.22823.1
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{D26F6F80-63EE-4081-A814-EF3DBABE24D9}"
 EndProject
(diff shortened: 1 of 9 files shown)


michaelvolz:
This looks promising. Any progress yet? I would love to use StructureMap with Asp.Net 5 / MVC 6.

ChrisTolmeijer:
I am also interested in the progress of this.

MrKevHunter:
@michaelvolz @ChrisTolmeijer You can just copy the code into your projects rather than waiting to see if it will be merged in, my understanding is that MS would rather not have the integration code in their lib

ceilidhboy:
There are several issues with leaving lifecycle as transient. First, transient as defined by ASP.NET doesn't quite mean the same as transient as defined by StructureMap; it actually maps to StructureMap's AlwaysUnique lifetime when used with a non-nested container. StructureMap's transient will create one instance per object graph whereas ASP.NET expects transient will always create unique instances.

Secondly, with a nested container, StructureMap's Transient means what ASP.NET calls Scoped.

So in order to match the ASP.NET scoping semantics, a bit more work needs to be done with the StructureMap adapter. Otherwise the behaviours will be subtly different which will result in unexpected results when StructureMap is used as a container instead of the default or other containers.

davidfowl:
We've moved all 3rd party containers out of this repository. It would be great to contribute this to the StructureMap itself. Thanks for the contribution!

michaelvolz:
just fyi: Here is a simple example how to integrate StructureMap 3 into Asp.Net 5:
http://josephwoodward.co.uk/2015/07/setting-up-structuremap-3-in-asp-net-mvc-6-vnext/

JoeMighty:
As mentioned in the blog post linked above (I'm the author), the code I'm using is a slightly updated (to take into consideration recent framework changes) version of this pull request as it was the only version I was able to get find and get working.

khellang:
There's now an "official" adapter that passes the test suite currently in this repository at https://github.com/structuremap/structuremap.dnx, with a NuGet package at https://www.nuget.org/packages/structuremap.dnx, if anyone comes here looking for it.

PR #236, opened 2015-05-13
Bugfix for non-generic GetRequiredServices

henkmollema:
The non-generic GetRequiredServices wasn't actually checking if the service exists.

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceProviderExtensions.cs
+++ b/src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceProviderExtensions.cs
@@ -82,7 +82,14 @@ namespace Microsoft.Framework.DependencyInjection
         public static IEnumerable<object> GetRequiredServices([NotNull] this IServiceProvider provider, [NotNull] Type serviceType)
         {
             var genericEnumerable = typeof(IEnumerable<>).MakeGenericType(new [] { serviceType });
-            return (IEnumerable<object>)provider.GetRequiredService(genericEnumerable);
+            var providers = (IEnumerable<object>)provider.GetRequiredService(genericEnumerable);
+            
+            if (!providers.Any())
+            {
+                throw new InvalidOperationException(Resources.FormatNoServiceRegistered(serviceType));
+            }
+
+            return providers;
         }
     }
 }
--- a/test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
+++ b/test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.AspNet.Testing;
 using Xunit;
 
 namespace Microsoft.Framework.DependencyInjection
@@ -11,10 +12,10 @@ namespace Microsoft.Framework.DependencyInjection
     public class ServiceProviderExtensionsTest
     {
         [Fact]
-        public void GetServicesReturnsAllServices()
+        public void GetRequiredServices_Returns_AllServices()
         {
             // Arrange
-            var serviceProvider = CreateFooServiceProvider(2);
+            var serviceProvider = CreateTestServiceProvider(2);
 
             // Act
             var services = serviceProvider.GetRequiredServices<IFoo>();
@@ -24,12 +25,12 @@ namespace Microsoft.Framework.DependencyInjection
             Assert.Contains(services, item => item is Foo2);
             Assert.Equal(2, services.Count());
         }
-        
+
         [Fact]
-        public void NonGeneric_GetRequiredServices_ReturnsAllServices()
+        public void NonGeneric_GetRequiredServices_Returns_AllServices()
         {
             // Arrange
-            var serviceProvider = CreateFooServiceProvider(2);
+            var serviceProvider = CreateTestServiceProvider(2);
 
             // Act
             var services = serviceProvider.GetRequiredServices(typeof(IFoo));
@@ -41,31 +42,74 @@ namespace Microsoft.Framework.DependencyInjection
         }
 
         [Fact]
-        public void GetServicesReturns_A_Service()
+        public void GetRequiredServices_Returns_SingleService()
         {
             // Arrange
-            var serviceProvider = CreateFooServiceProvider(1);
+            var serviceProvider = CreateTestServiceProvider(1);
 
             // Act
             var services = serviceProvider.GetRequiredServices<IFoo>();
 
             // Assert
-            Assert.Contains(services, item => item is Foo1);
-            Assert.Equal(1, services.Count());
+            var item = Assert.Single(services);
+            Assert.IsType<Foo1>(item);
         }
 
         [Fact]
-        public void ThrowsWhenNoServicesAreRegistered()
+        public void NonGeneric_GetRequiredServices_Returns_SingleService()
         {
             // Arrange
-            var serviceProvider = CreateFooServiceProvider(0);
+            var serviceProvider = CreateTestServiceProvider(1);
 
-            // Act + Assert
-            Assert.Throws<InvalidOperationException>(() => serviceProvider.GetRequiredServices<IFoo>());
+            // Act
+            var services = serviceProvider.GetRequiredServices(typeof(IFoo));
+
+            // Assert
+            var item = Assert.Single(services);
+            Assert.IsType<Foo1>(item);
         }
 
         [Fact]
-        public void ThrowsWhenNoServicesAreAvailable()
+        public void GetRequiredServices_Returns_CorrectTypes()
+        {
+            // Arrange
+            var serviceProvider = CreateTestServiceProvider(4);
+
+            // Act
+            var services = serviceProvider.GetRequiredServices(typeof(IAnotherFoo));
+
+            // Assert
+            foreach (var service in services)
+            {
+                Assert.IsAssignableFrom<IAnotherFoo>(service);
+            }
+            Assert.Equal(2, services.Count());
+        }
+
+        [Fact]
+        public void Throws_WhenNo_Services_AreRegistered()
+        {
+            // Arrange
+            var serviceProvider = CreateTestServiceProvider(0);
+
+            // Act + Assert
+            ExceptionAssert.Throws<InvalidOperationException>(() => serviceProvider.GetRequiredServices<IFoo>(),
+                $"No service for type '{typeof(IFoo)}' has been registered.");
+        }
+
+        [Fact]
+        public void Throws_WhenNo_Services_AreRegistered_NonGeneric()
+        {
+            // Arrange
+            var serviceProvider = CreateTestServiceProvider(0);
+
+            // Act + Assert
+            ExceptionAssert.Throws<InvalidOperationException>(() => serviceProvider.GetRequiredServices(typeof(IFoo)),
+                $"No service for type '{typeof(IFoo)}' has been registered.");
+        }
+
+        [Fact]
+        public void Throws_WhenNo_Services_AreAvailable()
         {
             // Arrange
             var serviceCollection = new ServiceCollection();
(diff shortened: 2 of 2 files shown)


henkmollema:
@davidfowl previous code was only working on the happy path, my mistake..

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceProviderExtensions.cs
```diff
@@ -62,14 +62,7 @@ public static T GetRequiredService<T>([NotNull] this IServiceProvider provider)
...
         {
-            var providers = provider.GetRequiredService<IEnumerable<T>>();
-
-            if (!providers.Any())
-            {
-                throw new InvalidOperationException(Resources.FormatNoServiceRegistered(typeof(T)));
-            }
-
-            return providers;
+            return (IEnumerable<T>)provider.GetRequiredServices(typeof(T));
```
davidfowl: That's less efficient id keep them separate .

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceProviderExtensions.cs
```diff
@@ -82,7 +75,14 @@ public static IEnumerable<T> GetRequiredServices<T>([NotNull] this IServiceProvi
         public static IEnumerable<object> GetRequiredServices([NotNull] this IServiceProvider provider, [NotNull] Type serviceType)
         {
             var genericEnumerable = typeof(IEnumerable<>).MakeGenericType(new [] { serviceType });
```
henkmollema: Does the performance of this matters? We can't benefit from the generic `IEnumerable<T>` as we previously did. This could be achieved using a private method which both methods call or just duplicate code (and tests).

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceProviderExtensions.cs
henkmollema: You beat me to it, I'll change it.

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceProviderExtensions.cs
davidfowl: Yes performance always matters. It's a 4 line method, just duplicate the check

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -24,21 +24,6 @@ public void GetServicesReturnsAllServices()
             Assert.Contains(services, item => item is Foo2);
             Assert.Equal(2, services.Count());
         }
-        
-        [Fact]
-        public void NonGeneric_GetRequiredServices_ReturnsAllServices()
-        {
```
davidfowl: Keep this test

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceProviderExtensions.cs
henkmollema: Agreed :+1:

henkmollema:
@davidfowl updated, including tests.

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -65,6 +79,16 @@ public void ThrowsWhenNoServicesAreRegistered()
         }
 
         [Fact]
+        public void ThrowsWhenNoServicesAreRegistered_NonGeneric()
+        {
+            // Arrange
+            var serviceProvider = CreateFooServiceProvider(0);
+
+            // Act + Assert
+            Assert.Throws<InvalidOperationException>(() => serviceProvider.GetRequiredServices(typeof(IFoo)));
```
pranavkm: Could you check the exception message here (so we know it's the `InvalidOperationException` you're throwing and not something else)?

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -77,6 +101,19 @@ public void ThrowsWhenNoServicesAreAvailable()
...
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddInstance<IEnumerable<IFoo>>(new List<IFoo>());
+
+            var serviceProvider = serviceCollection.BuildServiceProvider();
+
+            // Act + Assert
+            Assert.Throws<InvalidOperationException>(() => serviceProvider.GetRequiredServices(typeof(IFoo)));
+        }
+
         private IServiceProvider CreateFooServiceProvider(int count)
```
pranavkm: Can we rename this? `CreateTestServiceProvider` \ `CreateServiceProvider`?

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -55,6 +55,20 @@ public void GetServicesReturns_A_Service()
...
+        public void NonGeneric_GetServices_Returns_A_Service()
+        {
+            // Arrange
+            var serviceProvider = CreateFooServiceProvider(1);
+
+            // Act
+            var services = serviceProvider.GetRequiredServices(typeof(IFoo));
+
+            // Assert
+            Assert.Contains(services, item => item is Foo1);
```
pranavkm:
```C#
var item = Assert.Single(services);
Assert.IsType<Foo1>(item);
```

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -55,6 +55,20 @@ public void GetServicesReturns_A_Service()
         }
 
         [Fact]
+        public void NonGeneric_GetServices_Returns_A_Service()
```
pranavkm: Could we have tests that verify that multiple `IFoo` values are returned and we that we filter correctly - so that asking for `typeof(IDifferentFoo)` does not return `IFoo` interfaces?

pranavkm:
:watch:

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
henkmollema: `GetRequiredServices_Returns_AllServices` already tests if multiple values are returned.

henkmollema:
@pranavkm updated

cc @davidfowl

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -41,31 +42,87 @@ public void NonGeneric_GetRequiredServices_ReturnsAllServices()
...
-            var serviceProvider = CreateFooServiceProvider(0);
+            var serviceProvider = CreateTestServiceProvider(1);
+
+            // Act
+            var services = serviceProvider.GetRequiredServices(typeof(IFoo));
+
+            // Assert
+            var item = Assert.Single(services);
+            Assert.IsType<Foo1>(item);
+            Assert.Equal(1, services.Count());
```
pranavkm: You shouldn't need this - `Assert.Single` throws if the sequence has more than one element in it.

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -41,31 +42,87 @@ public void NonGeneric_GetRequiredServices_ReturnsAllServices()
...
+        {
+            // Arrange
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddInstance<IEnumerable<IFoo>>(new List<IFoo>());
+
+            var serviceProvider = serviceCollection.BuildServiceProvider();
+
+            // Act + Assert
+            ExceptionAssert.Throws<InvalidOperationException>(() => serviceProvider.GetRequiredServices<IFoo>(),
+                string.Format("No service for type '{0}' has been registered.", typeof(IFoo)));
```
pranavkm: This is fine as is, but you could write this as `$"No service for type '{typeof(IFoo)}' has been registered."`

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -41,31 +42,87 @@ public void NonGeneric_GetRequiredServices_ReturnsAllServices()
...
+        public void GetRequiredServices_Returns_CorrectTypes()
+        {
+            // Arrange
+            var serviceProvider = CreateTestServiceProvider(4);
+
+            // Act
+            var services = serviceProvider.GetRequiredServices(typeof(IAnotherFoo));
+
+            // Assert
+            Assert.True(services.All(s => s is IAnotherFoo));
```
pranavkm: Maybe use Assert.Collection here:

```C#
Assert.Collection(
   services, 
   Assert.IsType<IAnotherFoo>, 
   Assert.IsType<IAnotherFoo>);
```

This way the assertion includes the type that fails the check rather than seeing a false.

pranavkm:
:watch: for few remaining comments.

henkmollema:
@pranavkm updated again

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -41,31 +42,76 @@ public void NonGeneric_GetRequiredServices_ReturnsAllServices()
...
+            // Act
+            var services = serviceProvider.GetRequiredServices(typeof(IAnotherFoo));
+
+            // Assert
+            foreach (var service in services)
+            {
+                Assert.True(
+                    service is IAnotherFoo,
+                    $"Service is of type '{service.GetType()}'. Expected: '{typeof(IAnotherFoo)}'.");
+            }
```
henkmollema: Used a `foreach` here since the collection is `IEnumerable<object>`, so can't use `Assert.Collection()` here.

pranavkm:
:shipit:  from me. cc @Eilon

henkmollema:
@pranavkm seems like AppVeyor doesn't like the `Microsoft.AspNet.Test`. I use it for the `ExceptionAssert` class, but is it even allowed in project outside the `AspNet` namespace? I could change it to just manually comparing the message from the exception variable.

Tragetaschen:
You've only got a reference to `"Microsoft.AspNet.Testing": "1.0.0-*"` when compiling for the dnx451 framework, not for dnxcore50. The test on the other hand are run using CoreCLR (see line 268).

henkmollema:
@Tragetaschen nice catch, thanks!

Eilon:
Please squash the commits if you think you're good to go.

henkmollema:
@Eilon updated

henkmollema:
@Eilon what's the status on this?

Eilon:
@henkmollema we should be able to merge it this week.

henkmollema:
@Eilon any update?

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
Eilon: Is there no built in assertion for a type check?

test/Microsoft.Framework.DependencyInjection.Tests/project.json
```diff
@@ -3,6 +3,7 @@
     "dependencies": {
         "Microsoft.Framework.DependencyInjection": "1.0.0-*",
         "Microsoft.Framework.NotNullAttribute.Sources": { "version": "1.0.0-*", "type": "build" },
+        "Microsoft.AspNet.Testing": "1.0.0-*",
```
Eilon: Sort

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
dougbu: Exact type: `Assert.IsType<T>(service)`
Compatible type: `Assert.IsAssignableFrom<T>(service)`

henkmollema:
@Eilon @dougbu updated

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceProviderExtensions.cs
```diff
@@ -82,7 +82,14 @@ public static IEnumerable<T> GetRequiredServices<T>([NotNull] this IServiceProvi
...
-            return (IEnumerable<object>)provider.GetRequiredService(genericEnumerable);
+            var providers = (IEnumerable<object>)provider.GetRequiredService(genericEnumerable);
+            
+            if (!providers.Any())
```
Eilon: This just in, from the world of premature optimization...

What perf risks are there from calling `.Any()` on this? I have this fear that common implementations of `IServiceProvider` might do costly work every time the result is enumerated. And with this change, the work might be done twice. I'm figuring this method here is called a zillion times per request, so that's why I have this concern.

henkmollema: I was wondering the same thing. But since the generic `GetRequiredServices<T>` also uses `Any()`, I figured it was okay.

We could already enumerate the `providers` collection, check if `Count > 0` and return that `List<object>`. That should avoid the multiple enumerations since it's already a list, right?

Perhaps @davidfowl can help?

Eilon: I'm probably fine with this as-is, but maybe log a bug to look at that in the future. Ultimately we'd have to run this in a profiler to see if this even matters.

henkmollema: @Eilon I'll leave it like this than. Rebased.

henkmollema:
Looks like it's messed up atm, will fix in a sec.

henkmollema:
Fixed and rebased

PR #239, opened 2015-05-13
Correct version for EF in README.md

henkmollema:
Entity Framework is on version 7, not 5. Also, Entity Framework looks better than EF imo.


Eilon:
Duh, nice catch!

PR #249, opened 2015-06-17
Add a TryAddMultiRegistration variant

rynowak:
We're using something like this in MVC to avoid repeatedly adding our
options setups when called twice. This makes AddMvcCore() idempotent,
which is pretty important to make our more complex configuration scenarios
more straightforward.

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceCollectionExtensions.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.Framework.DependencyInjection.Abstractions;
 using Microsoft.Framework.Internal;
 
 namespace Microsoft.Framework.DependencyInjection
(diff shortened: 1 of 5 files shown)


davidfowl:
This one is a little weird.

rynowak:
Needed for options-setups and other services that allow multiples. MVC will have a copy of this code x20 if you don't like it :laughing:.

Try imaging this with a different name and then tell me if it's wierd.

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceCollectionExtensions.cs
```diff
@@ -184,6 +185,53 @@ public static bool TryAddSingleton<TService>([NotNull] this IServiceCollection c
...
+        /// </returns>
+        /// <remarks>
+        /// Use <see cref="TryAddMultiRegistration"/> when registing a service implementation of a service type that
+        /// supports multiple registrations of the same service type. Using
+        /// <see cref="Add(IServiceCollection, ServiceDescriptor)"/> is not idempotent and can add duplicate
+        /// <see cref="ServiceDescriptor"/> instances if called twice. Using
+        /// <see cref="TryAddMultiRegistration(IServiceCollection, ServiceDescriptor)"/> will prevent registration
+        /// of multiple implementation types.
+        /// </remarks>
+        public static bool TryAddMultiRegistration(
```
davidfowl: Ok can we pick another name? Mult and Registration are never seen before weird.

rynowak: What's the suggestion? `TryAddImplementation`?

HaoK: TryAddUnique?  TryAddMultiRegistration does sound pretty weird

HaoK: Or TryAddIfUnique reads a bit better

mikary: Naming seems weird since we don't have a real concept of adding to an IEnumerable<TService> distinguished from replacing a previous definition of a service type.

PR #252, opened 2015-06-26
Add TryAddEnumerable overload and Add implementation factory overloads

mikary:
These changes will help make AddEntityFramework() idempotent

@rynowak This change removes the exception thrown by TryAddEnumerable, will this break any scenarios for MVC?

/cc @davidfowl

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection.Abstractions/Resources.resx
+++ b/src/Microsoft.Framework.DependencyInjection.Abstractions/Resources.resx
@@ -133,12 +133,16 @@
     <value>Unable to locate suitable constructor for type '{0}'. Ensure the type is concrete and all parameters are accepted by a constructor.</value>
     <comment>{0} = service type</comment>
   </data>
+  <data name="NoImplementation" xml:space="preserve">
+    <value>Service type '{0}' has been registered without an implementation type.</value>
+    <comment>{0} = service type</comment>
+  </data>
   <data name="NoServiceRegistered" xml:space="preserve">
     <value>No service for type '{0}' has been registered.</value>
     <comment>{0} = service type</comment>
   </data>
-  <data name="TryAddEnumerable_ImplementationTypeMustBeSet" xml:space="preserve">
-    <value>The {0} must have the {1} property set to a non-null value.</value>
-    <comment>{0} = ServiceDescriptor, {1} = ImplementationType</comment>
+  <data name="TryAddIndistinguishableTypeToEnumerable" xml:space="preserve">
+    <value>Implementation type cannot be '{0}' because it is indistinguishable from other services registered for '{1}'.</value>
+    <comment>{0} = implementation type, {1} = service type</comment>
   </data>
 </root>
--- a/src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceCollectionExtensions.cs
@@ -46,16 +46,13 @@ namespace Microsoft.Framework.DependencyInjection
         /// </summary>
         /// <param name="collection">The <see cref="IServiceCollection"/>.</param>
         /// <param name="descriptor">The <see cref="ServiceDescriptor"/>.</param>
-        /// <returns><c>true</c> if the <paramref name="descriptor"/> was added; otherwise <c>false</c>.</returns>
-        public static bool TryAdd([NotNull] this IServiceCollection collection,
+        public static void TryAdd([NotNull] this IServiceCollection collection,
                                   [NotNull] ServiceDescriptor descriptor)
         {
             if (!collection.Any(d => d.ServiceType == descriptor.ServiceType))
             {
                 collection.Add(descriptor);
-                return true;
             }
-            return false;
         }
 
         /// <summary>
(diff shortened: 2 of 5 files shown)


src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceDescriptor.cs
```diff
@@ -72,6 +72,24 @@ private ServiceDescriptor(Type serviceType, ServiceLifetime lifetime)
         /// <inheritdoc />
         public Func<IServiceProvider, object> ImplementationFactory { get; }
 
+        public Type GetImplementationType()
```
mikary: The name for ```GetImplementationType()``` is kind of awful but ```ImplementaitonType``` is already taken with different semantics.

Might also be better as an auto property set on construction

rynowak: Why not just put this in the extensions class as a private method rather than create a new public surface area?

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceDescriptor.cs
```diff
@@ -72,6 +72,24 @@ private ServiceDescriptor(Type serviceType, ServiceLifetime lifetime)
...
+            {
+                return ImplementationType;
+            }
+            else if (ImplementationInstance != null)
+            {
+                return ImplementationInstance.GetType();
+            }
+            else if (ImplementationFactory != null)
+            {
+                return ImplementationFactory.GetType().GenericTypeArguments[1];
```
rynowak: This seems very fragile to do without some error checking

halter73: `ImplementationFactory` is always a `Func<IServiceProvider, object>`, so this should be safe.

I wonder if we should throw an exception if the returned type would be `object`.

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceDescriptor.cs
halter73: :+1:

mikary:
Quick example of the pattern we are using for the new `AddTransient`, `AddScoped` and `AddSingleton` overloads. After these changes, the following will be idempotent and not collide with other types in the collections

```cs
serviceCollection.TryAddEnumerable(new ServiceCollection()
    .AddScoped<IEntityStateListener, INavigationFixer>(p => p.GetService<INavigationFixer>())
    .AddScoped<IForeignKeyListener, INavigationFixer>(p => p.GetService<INavigationFixer>())
    .AddScoped<INavigationListener, INavigationFixer>(p => p.GetService<INavigationFixer>())
    .AddScoped<IKeyListener, INavigationFixer>(p => p.GetService<INavigationFixer>())
    .AddScoped<IPropertyListener, IChangeDetector>(p => p.GetService<IChangeDetector>()));
```

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceDescriptor.cs
pranavkm: Could you throw in a `Debug.Assert(ImplementationFactory.GetType().GenericTypeArguments.Length == 2)`?

mikary:
Updated with changes from PR feedback.

Went in a slightly different direction with the `object` error message than @halter73 suggested, but I think it's more appropriate for the intended use.

src/Microsoft.Framework.DependencyInjection.Abstractions/Resources.resx
```diff
@@ -133,12 +133,20 @@
...
+  </data>
   <data name="NoServiceRegistered" xml:space="preserve">
     <value>No service for type '{0}' has been registered.</value>
     <comment>{0} = service type</comment>
   </data>
-  <data name="TryAddEnumerable_ImplementationTypeMustBeSet" xml:space="preserve">
-    <value>The {0} must have the {1} property set to a non-null value.</value>
-    <comment>{0} = ServiceDescriptor, {1} = ImplementationType</comment>
+  <data name="TryAddIndistinguishableTypeToEnumerable" xml:space="preserve">
+    <value>Unable to add service to Enumerable because the implementation type '{0}' is indistinguishable from the service type '{1}'.</value>
```
rynowak: Any code-like names should also be parameters to the format string to avoid overlocalization.

halter73: How about "Implementation type cannot be '{0}' because it is indistinguishable from other services registered for '{1}'."

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceCollectionExtensions.cs
```diff
@@ -210,28 +208,72 @@ public static bool TryAddSingleton<TService>([NotNull] this IServiceCollection c
...
-            if (services.Any(d =>
-                d.ServiceType == descriptor.ServiceType &&
-                d.ImplementationType == descriptor.ImplementationType))
+            foreach(var service in services.Where(s => s.ServiceType == serviceType))
             {
-                return false;
+                var serviceImplementationType = service.GetImplementationType();
+
+                if (serviceImplementationType == typeof(object)
+                    || serviceImplementationType == serviceType)
```
halter73: I don't think we need to be this protective.

mikary:
More updates from feedback

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceCollectionExtensions.cs
```diff
@@ -307,6 +339,15 @@ public static IServiceCollection AddTransient<TService>([NotNull] this IServiceC
             return services.AddTransient(typeof(TService), implementationFactory);
         }
 
+        public static IServiceCollection AddTransient<TService, TImplementation>(
+            [NotNull] this IServiceCollection services,
+            [NotNull] Func<IServiceProvider, TImplementation> implementationFactory)
+            where TService : class
+            where TImplementation : class
```
halter73: `AddTransient<TService, TImplementation>` has a `where TImplementation : TService` constraint. Maybe we should do the same here. Same for scoped and singleton.

`TryAdd*<TService, TImplementation>` don't seem to have any constraints, so maybe we should add them there too.

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceDescriptor.cs
```diff
@@ -83,6 +106,14 @@ public static ServiceDescriptor Transient([NotNull] Type service, [NotNull] Type
             return Describe(service, implementationType, ServiceLifetime.Transient);
         }
 
+        public static ServiceDescriptor Transient<TService, TImplementation>(
+            [NotNull] Func<IServiceProvider, TImplementation> implementationFactory)
+            where TService : class
+            where TImplementation : class
```
halter73: I think we should also use stronger generic constraints in this class where possible.

test/Microsoft.Framework.DependencyInjection.Tests/ServiceCollectionExtensionTest.cs
```diff
@@ -276,87 +285,154 @@ public void TryAddWithEnumerableReturnsFalseIfNoneAdded()
...
-            Assert.Single(collection, sd => sd.ImplementationType == typeof(IFakeService));
+            var d = Assert.Single(collection);
+            Assert.Equal(expectedServiceType, d.ServiceType);
+            Assert.Equal(expectedImplementationType, d.GetImplementationType());
+            Assert.Equal(expectedLifetime, d.Lifetime);
         }
 
         [Fact]
-        public void TryAddEnumerable_AllowsMultipleRegistrations_WithFactory()
+        public void TryAddEnumerable_WithEnumerable_ReturnsTrueIfAnyAdded()
```
halter73: Can you explain why we want `TryAddEnumerable` to return true if any descriptor is added as opposed to only returning true if all descriptors are added? How does calling code use this return value?

test/Microsoft.Framework.DependencyInjection.Tests/ServiceCollectionExtensionTest.cs
```diff
@@ -276,87 +285,154 @@ public void TryAddWithEnumerableReturnsFalseIfNoneAdded()
...
+                    { ServiceDescriptor.Scoped(serviceType, s => new FakeService()), serviceType, objectType },
+
+                    { ServiceDescriptor.Singleton<IFakeService>(s => new FakeService()), serviceType, serviceType },
+                    { ServiceDescriptor.Singleton(serviceType, s => new FakeService()), serviceType, objectType },
+                };
+            }
         }
 
-        [Fact]
-        public void TryAddEnumerable_ThrowsForNonType()
```
halter73: Maybe we add a test similar to this one that was removed for when `ImplementationType`, `ImplementationInstance` and `ImplementionFactory` are all null.

test/Microsoft.Framework.DependencyInjection.Tests/ServiceCollectionExtensionTest.cs
```diff
@@ -64,25 +65,33 @@ public static TheoryData AddImplementationFactoryData
...
                 var serviceType = typeof(IFakeService);
+                var implementationType = typeof(FakeService);
+                var objectType = typeof(object);
 
-                return new TheoryData<Action<IServiceCollection>, ServiceLifetime>
+                return new TheoryData<Action<IServiceCollection>, Type, Type, ServiceLifetime>
                 {
-                    { collection => collection.AddTransient<IFakeService>(_factory), ServiceLifetime.Transient },
-                    { collection => collection.AddTransient(serviceType, _factory), ServiceLifetime.Transient },
+                    { collection => collection.AddTransient(serviceType, s => new FakeService()),serviceType, objectType, ServiceLifetime.Transient },
```
halter73: Nit: Ctrl+K+D

halter73:
I'm still skeptical that we should be using the ServiceCollection in this way.

While I understand the desire to prevent duplicate services from being registered, looking at the implementation type of the registered service might not be the best way to prevent this from happening. What if I register two different service implementations that happen to have the same type (perhaps they were just configured differently)?

This wasn't as much of an issue when you could only register `ImplentationType`s using `TryAddEnumerable`, because then to services with the same `ImplementationType` couldn't be configured differently. Now they can.

Do we expect end users to call `TryAddEnumerable`? If so, I think this could be very confusing considering there is already a `TryAdd` method that won't add a service if there is any service already in the collection with the same `ServiceType`. `TryAddEnumerable`'s behavior is quite different, but I think the average user would expect it to behave similarly to `TryAdd`. If we don't expect end users to call `TryAddEnumerable`, it probably doesn't belong in ServiceCollectionExtensions.

I think this PR does a good job of extending the `TryAddEnumerable` functionality to `ServiceDescriptor`s without `ImplemenationType`s, but I'm not sure this is something we really want or need. Maybe I just don't understand the use case well enough yet.

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceDescriptor.cs
mikary: Seems reasonable to add generic constraints for TImplementation wherever possible. I'm not sure why these weren't included in the first place (ANI?) but I'll add them in.

test/Microsoft.Framework.DependencyInjection.Tests/ServiceCollectionExtensionTest.cs
mikary: I was going for consistency with the existing `TryAdd` implementation. For the EF use case, the calling code ignores the return value either way.

mikary:
It is a difficult pattern to describe / document, but seems to be useful for both MVC and EF. Perhaps it would be best to have this behavior in a different namespace or just move it into the product code. Alternatively, it might be best to revisit how EF (and other products) use the `ServiceCollection`, but that discussion is well beyond the scope of this PR.

halter73:
After discussion, we decided to create [a new issue](https://github.com/aspnet/DependencyInjection/issues/254) to put the extension methods used primarily by frameworks (e.g. `TryAdd`  and `TryAddEnumerable`) into a separate namespace so we don't clutter intellisense. Since this will probably require changes to a lot of other repos, it's probably better to work on the new issue in another PR.

:shipit:

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceCollectionExtensions.cs
```diff
@@ -232,6 +231,39 @@ public static bool TryAddSingleton<TService>([NotNull] this IServiceCollection c
...
+        /// </returns>
+        /// <remarks>
+        /// Use <see cref="TryAddEnumerable"/> when registing a service implementation of a service type that
+        /// supports multiple registrations of the same service type. Using
+        /// <see cref="Add(IServiceCollection, ServiceDescriptor)"/> is not idempotent and can add duplicate
+        /// <see cref="ServiceDescriptor"/> instances if called twice. Using
+        /// <see cref="TryAddEnumerable(IServiceCollection, ServiceDescriptor)"/> will prevent registration
+        /// of multiple implementation types.
+        /// </remarks>>
+        public static bool TryAddEnumerable(
```
halter73: I'm really tempted to make this method and the `TryAdd` equivalent void. If no one is really using a return value, we shouldn't return something that isn't intuitive.

mikary:
I took a look through MVC's repo and it looks like the return value of `TryAddEnumerable` is never used, so it makes sense to take out in this PR. Probably makes more sense to address the return value of `TryAdd` in #254

halter73:
I would make TryAdd void in this PR. Just like you are changing the generic constraints in methods you didn't change in this PR. Might as well make it consistent since it doesn't require changing any other repos.

mikary:
As long as we're looking at the type constraints, do we want to be more consistent about ensuring `TService` and `TImplementation` are reference types? About 1 in 3 of the extensions have this constraint right now.

Issue #264, opened 2015-07-27
Remove non-official DI samples


Eilon:
Remove Ninject and Autofac.

PR #266, opened 2015-07-28
* Removing non-official DI samples

pranavkm:
* Switching to Autofac's nightly feed for unit tests.

Fixes https://github.com/aspnet/DependencyInjection/issues/264


pranavkm:
cc @Eilon

Eilon:
:shipit:

test/Microsoft.Framework.DependencyInjection.Tests/project.json
```diff
@@ -1,19 +1,17 @@
 {
     "version": "0.1-alpha-*",
     "dependencies": {
+        "Autofac.Framework.DependencyInjection": "4.0.0-*",
         "Microsoft.AspNet.Testing": "1.0.0-*",
         "Microsoft.Framework.DependencyInjection": "1.0.0-*",
-        "Microsoft.Framework.NotNullAttribute.Sources": { "version": "1.0.0-*", "type": "build" },
+        "Microsoft.Framework.NotNullAttribute.Sources": {
```
davidfowl: No bad formatting :smile:

davidfowl:
Do we still need to run unit tests for autofac?

test/Microsoft.Framework.DependencyInjection.Tests/project.json
pranavkm: VS 2015 RTM defaults to this. It's a pain to have it switch.

rynowak: We should not fight vs on formatting, it's more work for everyone.

Eilon:
@davidfowl we want to keep them at least for now to make sure we don't break anything.

pranavkm:
@davidfowl the plan was to run tests against the Autofac nightlies so they act as a smoke test. Mvc has tests that run against the Autofac on nuget.org so this should give us reasonable coverage.

pranavkm:
Waiting on this one until https://github.com/aspnet/Mvc/pull/2869 goes in.

alexmg:
I temporarily pulled the unit tests from this project into Autofac for a sanity check against what is essentially the specification. It would be nice to have a dependency that I could pull in that contains these tests, allowing only the `AutofacContainerTests` that returns the populated container as `IServiceProvider` to be present in our project. Something like a `Microsoft.Framework.DependencyInjection.Specification` or `Microsoft.Framework.DependencyInjection.Tests` package for DI container authors.

davidfowl:
@alexmg Great idea. We should make this a package.

Eilon:
Yeah that would be cool. Logged https://github.com/aspnet/DependencyInjection/issues/270 to track this.

PR #268, opened 2015-07-29
Adding unit tests for consuming services without implementations

### LINKED ISSUE #210: NullRef in ctor invocation
Ran into this when I accidentally did `services.AddTransient<IFoo, IFoo>()` and then had another class ctor `public Ctor(IFoo foo)`.  DI throws a Null ref deep inside of trying to invoke the ctor instead of a useful error message that was swallowed (cannot create an instance of an interface)

pranavkm:
Fixes #210


pranavkm:
cc @Eilon

Eilon:
:shipit:

Issue #241, opened 2015-05-17
ServiceProviderExtensions missing GetServices and GetServices<T> extension method


Tasteful:
The `ServiceProviderExtensions` contains methods for `GetRequiredServices`/`GetRequiredServices<T>`. Equivalent method for `GetServices`/`GetServices<T>` without the `providers.Any()` check that always returning an empty `IEnumerable<T>` when non existings instances is registered in the container.

henkmollema:
+1

Eilon:
The final set of methods we want to have are:

```c#
// Keep these:
GetService() // comes from IServiceProvider interface
GetService<T>()

GetRequiredService()
GetRequiredService<T>()

// Add these:
GetServices()
GetServices<T>()

// Delete these:
GetRequiredServices()
GetRequiredServices<T>()
```

henkmollema:
@Eilon I could look at this if you'd like.

henkmollema:
Why drop the `GetRequiredServices` btw?

rynowak:
> Why drop the GetRequiredServices btw?

We couldn't think of a use-case where you need 1 or more of something. Given the number of methods that already exist we wanted to make it easier to pick the one you want.

henkmollema:
When applying the mediator pattern, I actually register multiple implementations of `INotificationHandler<TNotification>` and retrieve them using `GetRequiredServices()`.I'll be fine with `GetServices`, however.

Is that going to be implemented using `GetRequiredService()` or just with `GetService()`?

pranavkm:
`GetServices` would do what `GetRequiredServices` does without the `Any` check:

```c#
public static IEnumerable<T> GetRequiredServices<T>([NotNull] this IServiceProvider provider)
 => provider.GetRequiredService<IEnumerable<T>>();
```

The plan is to do `Any` check on the consumer if required.

pranavkm:
Fixed in e490facfd40d46ed73b8c601c9c731f4e251dead

PR #267, opened 2015-07-28
Update .travis.yml

davidfowl:
Removed mono 3.12.0 restriction

PR #269, opened 2015-07-29
Replace GetServices with GetRequiredServices

henkmollema:
GetServices returns an emty array or list rather than throwing an exception when there are no services registered. The consumer should check with `Any()` if there are any services.

Added tests for every method in `ServiceProviderExtensions`

#241

### DIFF
--- a/test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
+++ b/test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
@@ -12,13 +12,48 @@ namespace Microsoft.Framework.DependencyInjection
     public class ServiceProviderExtensionsTest
     {
         [Fact]
-        public void GetRequiredServices_Returns_AllServices()
+        public void GetService_Returns_CorrectService()
+        {
+            // Arrange
+            var serviceProvider = CreateTestServiceProvider(1);
+
+            // Act
+            var service = serviceProvider.GetService<IFoo>();
+
+            // Assert
+            Assert.IsType<Foo1>(service);
+        }
+
+        [Fact]
+        public void GetRequiredService_Throws_WhenNoServiceRegistered()
+        {
+            // Arrange
+            var serviceProvider = CreateTestServiceProvider(0);
+
+            // Act + Assert
+            ExceptionAssert.Throws<InvalidOperationException>(() => serviceProvider.GetRequiredService<IFoo>(),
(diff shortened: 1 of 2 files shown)


henkmollema:
Build fails because of 'unknown header' errors:
```
Warning: DownloadPackageAsync: https://www.nuget.org/api/v2/package/System.Runtime/4.0.0-beta-23109
  Unknown header: 1701869908
```

pranavkm:
Think that has to something to do with https://github.com/aspnet/DependencyInjection/pull/267

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -70,73 +105,83 @@ public void NonGeneric_GetRequiredServices_Returns_SingleService()
...
+            Assert.Empty(services);
+            Assert.IsType<List<IFoo>>(services);
         }
 
         [Fact]
-        public void Throws_WhenNo_Services_AreAvailable_NonGeneric()
+        public void NonGeneric_GetServices_WithBuildServiceProvider_Returns_EmptyList_WhenNoServicesAvailable()
         {
             // Arrange
             var serviceCollection = new ServiceCollection();
```
pranavkm: Could we add a variant of this test (and the one above) where `List<Foo>` is not registered? Would be a better verification for the scenario where no services of a type are registered and we invoke `GetServices` on it.

pranavkm:
:watch: for the two additonal tests.

test/Microsoft.Framework.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
henkmollema: Isn't that covered by `GetServices_Returns_EmptyArray_WhenNoServicesAvailable`? That creates an empty service provider and checks if the result is an empty array `IFoo[]`.

henkmollema: @pranavkm what's your opinion on this?

pranavkm: Yeah, I didn't see that test earlier. This looks good. I'll get it in today.

pranavkm:
Merged in e490facfd40d46ed73b8c601c9c731f4e251dead

PR #271, opened 2015-08-01
Open Generic services don't work with anything but implementation types

### LINKED ISSUE #257: Open Generic services don't work with anything but implementation types
The generic service registration assumes an implementation type and explodes with a null ref if one wasn't provided. We should either throw up front or support this in some way. Supporting an implementation factory would mean that the callback would need access to the closed generic.

/cc @lodejard

pranavkm:
Fixes #257

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceDescriptor.cs
+++ b/src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceDescriptor.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Reflection;
 using Microsoft.Framework.DependencyInjection.Abstractions;
 using Microsoft.Framework.Internal;
 
--- a/DependencyInjection.sln
+++ b/DependencyInjection.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.22803.0
+VisualStudioVersion = 14.0.23107.0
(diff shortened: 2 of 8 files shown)


pranavkm:
cc @davidfowl \ @Eilon

src/Microsoft.Framework.DependencyInjection.Abstractions/ServiceDescriptor.cs
```diff
@@ -22,6 +23,14 @@ public class ServiceDescriptor
                                  ServiceLifetime lifetime)
             : this(serviceType, lifetime)
         {
+            if (serviceType.GetTypeInfo().IsGenericTypeDefinition &&
```
pranavkm: Thoughts on verifying it in the `ServiceDescriptor`? Seemed like the one place to all the registrations get funneled to.

Eilon: I don't really know any better, but that sounds fine to me.

Eilon:
:shipit:

davidfowl:
Not sure this fix is in the right place. What if this is supported on other DI implementations? The thing that throws the exception today is the default implementation. The descriptor is just metadata. This is where it should throw from:

https://github.com/aspnet/DependencyInjection/blob/dev/src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceTable.cs#L27

pranavkm:
Fair enough.

pranavkm:
Updated

Eilon:
Ok still looks good to me... :shipit:

PR #273, opened 2015-08-04
Ambiguous constructor matching is not working well

### LINKED ISSUE #69: Ambiguous constructor matching is not working well
This is a multi-issue bug but since the issues are extermly related, I'll open only one. 

This bug is the result of an investigation around bug  https://github.com/aspnet/Identity/issues/50 . The RoleManager<T> class has two constructors both with parameters.

~~~C#
public class RoleManager<TRole> : IDisposable where TRole : class
{
    public RoleManager(IRoleStore<TRole> store);
    public RoleManager(IServiceProvider services);
~~~

When one tries to resolve an instance, an exception saying that there is no parameter less constructor is thrown.

The out-of-the-box container has a few issues resolving objects with multiple constructors.

First, there is no way to specify which constructor should be used when injecting. Most DI framework allow you to mark a constructor with an attribute so that no ambiguity exists.

Second, the logic that we have today to resolve multi-constructor object instances seems incorrect. See the comments that I made on code.

From ActivatorUtilities.cs

~~~C#
public static Func<IServiceProvider, object> CreateFactory(Type type)
{
    if (type == null)
    {
        throw new ArgumentNullException("type");
    }

    // Why injectable constructors are only those that are public with parameters?
    // A parameter less constructor is injectable too.
    ConstructorInfo[] constructors = type.GetTypeInfo()
        .DeclaredConstructors
        .Where(IsInjectable)
        .ToArray();

    // If there are two constructors, one with params
    // and one without, why do we default to the one
    // with parameters?
    if (constructors.Length == 1)
    {
        ParameterInfo[] parameters = constructors[0].GetParameters();
        return services =>
        {
            var args = new object[parameters.Length];
            for (int index = 0; index != parameters.Length; ++index)
            {
                args[index] = services.GetService(parameters[index].ParameterType);
            }
            return Activator.CreateInstance(type, args);
        };
    }

    // We only use the default constructor if there is no other constructor
    // with parameters or more than one with parameters. Why?
    return _ => Activator.CreateInstance(type);
}

private static bool IsInjectable(ConstructorInfo constructor)
{
    return constructor.IsPublic && constructor.GetParameters().Length != 0;
}

~~~

My 2c on how I would expect this to work:
1. If there is only one public constructor => use that one and resolve any parameters
2. If there are more than one public constructors => throw an ambiguous match exception
3. If there are more than one public constructors but one of them is marked somehow as injectable (attribute?) => use that one

Another idea that I don't like too much but might work:
See that RoleManager<T> has a constructor that takes an IServiceProvider. Maybe we should default to that in case of ambiguities?

pranavkm:
Fixes #69

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceTable.cs
+++ b/src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceTable.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Reflection;
 
 namespace Microsoft.Framework.DependencyInjection.ServiceLookup
@@ -26,13 +27,24 @@ namespace Microsoft.Framework.DependencyInjection.ServiceLookup
                 var serviceTypeInfo = descriptor.ServiceType.GetTypeInfo();
                 if (serviceTypeInfo.IsGenericTypeDefinition)
                 {
-                    if (descriptor.ImplementationType == null ||
-                        !descriptor.ImplementationType.GetTypeInfo().IsGenericTypeDefinition)
+                    var implementationTypeInfo = descriptor.ImplementationType?.GetTypeInfo();
+
+                    if (implementationTypeInfo == null ||
+                        !implementationTypeInfo.IsGenericTypeDefinition)
                     {
                         throw new ArgumentException(
                             Resources.FormatOpenGenericServiceRequiresOpenGenericImplementation(
                                 descriptor.ServiceType),
                             nameof(descriptors));
+                    }
+
+                    if (implementationTypeInfo.IsAbstract ||
+                        implementationTypeInfo.IsInterface)
+                    {
+                        throw new ArgumentException(
+                            Resources.FormatTypeCannotBeActivated(
+                                descriptor.ImplementationType,
+                                descriptor.ServiceType));
                     }
 
                     Add(descriptor.ServiceType, new GenericService(descriptor));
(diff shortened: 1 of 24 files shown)


davidfowl:
Do we also need to fix https://github.com/aspnet/Common/tree/dev/src/Microsoft.Framework.ActivatorUtilities.Sources?

Eilon:
Couldn't tell if the test scenario of a ctor with an `IEnumerable<TService>` and other more complex DI scenarios are covered here... are they?

pranavkm:
@davidfowl Since we can't do the service registration lookup as part of `ActivatorUtilities`, the plan was to not change its current behavior.

pranavkm:
@Eilon Doesn't cover `IEnumerable<TService>` - I'll add a few combinations tests that deal with it + generic types. I'll work under the assumption that we are in no way special casing collection types. For instance we won't treat these as overlapping:

* `public Foo(List<IService> services)`
* `public Foo(IList<IService> services, int count)`

pranavkm:
Updated

Eilon:
:shipit:

src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceTable.cs
```diff
@@ -47,6 +59,19 @@ public ServiceTable(IEnumerable<ServiceDescriptor> descriptors)
                 }
                 else
                 {
+                    Debug.Assert(descriptor.ImplementationType != null);
+                    var implementationTypeInfo = descriptor.ImplementationType.GetTypeInfo();
```
pranavkm: cc @Eilon. Rebasing on dev was causing one of the more recently added tests to fail. The issue was that we were attempting to initialize one of the constructors for `List<>` which then blew up with an exception about being unable to create a void type array. It seems simpler to perform a pre-check as part of registration so we never have to deal with these cases later on.

Eilon: Which test was failing? Not sure I understand the scenario... though, the code in the `else` I think is fine...

Eilon: Talked in person; this is fine.

Eilon:
:shipit:

PR #279, opened 2015-08-19
* Removing NotNull attribute (that was causing warnings) from test

pranavkm:
* Add warningAsErrors to source projects.


davidfowl:
Why ?

pranavkm:
It prints a warning every time you build. If we turn `warningAsErrors` on in the test, it fails.

davidfowl:
:shipit:

PR #283, opened 2015-08-28
Fix #282 - Remove allocating lambda in GetService

### LINKED ISSUE #282: [PERF] Remove lambda allocation in ServiceProvider.GetService
https://github.com/aspnet/DependencyInjection/blob/dev/src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs#L52

This lamdba is allocating about 6mb across 10000 requests of a simple MVC app. We could easily make this 0 by turning the lambda into an instance method and caching a delegate to it.

rynowak:


src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
```diff
@@ -49,17 +56,19 @@ internal ServiceProvider(ServiceProvider parent)
...
+            if (callSite != null)
             {
-                var callSite = GetServiceCallSite(key, new HashSet<Type>());
-                if (callSite != null)
-                {
-                    return RealizeService(_table, key, callSite);
-                }
-                return _ => null;
-            });
+                return RealizeService(_table, serviceType, callSite);
```
davidfowl: Doesn't this allocate too?

rynowak: Yes, but only once, then it's cached by `ServiceTable`. The problem described in the bug is one endemic to `GetOrAdd` - you usually have to allocate just to call the function.

pranavkm:
:shipit:

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
davidfowl: I see

PR #284, opened 2015-09-01
Use earliest compatible dependencies

bricelam:
Enables our packages to work on more platforms with more libraries


bricelam:
DNX isn't handling this well. Will reconsider later.

PR #289, opened 2015-09-14
[Design] Reduce DI Allocations

### LINKED ISSUE #288: Reduce DI Allocations
For 2000 plantext requests DI allocates 16,009 objects:

image

Clocking in at 788,304 bytes:

image

Which would mean if you were performing 1M rps the GC would have to flush 8,004,500 objects and 394 Mbytes per second, 788 Mbytes per second for 2M rps and 2.76GBytes (56M objects) per second for 7M rps (~10GbE saturation)

benaadams:
Resolves #288

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
+++ b/src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
@@ -9,12 +9,8 @@ namespace Microsoft.Framework.DependencyInjection.ServiceLookup
     {
         private readonly ServiceProvider _scopedProvider;
         private readonly ServiceScopeFactory _parentFactory;
-
-        public ServiceScope(ServiceProvider scopedProvider)
-        {
-            _scopedProvider = scopedProvider;
-        }
-        internal ServiceScope(ServiceProvider scopedProvider, ServiceScopeFactory parentFactory)
+        
+        public ServiceScope(ServiceProvider scopedProvider, ServiceScopeFactory parentFactory)
         {
             _scopedProvider = scopedProvider;
             _parentFactory = parentFactory;
--- a/src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
+++ b/src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
@@ -23,8 +23,7 @@ namespace Microsoft.Framework.DependencyInjection.ServiceLookup
 
         internal void Reset()
         {
-            // Double wash
-            _scopedProvider.Dispose();
+            _scopedProvider.Reset();
         }
 
         public void Dispose()
--- a/src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
+++ b/src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
@@ -8,10 +8,16 @@ namespace Microsoft.Framework.DependencyInjection.ServiceLookup
     internal class ServiceScope : IServiceScope
     {
         private readonly ServiceProvider _scopedProvider;
+        private readonly ServiceScopeFactory _parentFactory;
 
         public ServiceScope(ServiceProvider scopedProvider)
         {
             _scopedProvider = scopedProvider;
+        }
+        internal ServiceScope(ServiceProvider scopedProvider, ServiceScopeFactory parentFactory)
+        {
+            _scopedProvider = scopedProvider;
+            _parentFactory = parentFactory;
         }
 
         public IServiceProvider ServiceProvider
@@ -19,9 +25,20 @@ namespace Microsoft.Framework.DependencyInjection.ServiceLookup
             get { return _scopedProvider; }
         }
 
+        internal void Reset()
+        {
+            // Double wash
+            _scopedProvider.Dispose();
+        }
+
         public void Dispose()
         {
             _scopedProvider.Dispose();
+
+            if (_parentFactory != null)
+            {
+                _parentFactory.PoolScope(this);
+            }
         }
     }
 }
--- a/src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScopeFactory.cs
+++ b/src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScopeFactory.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Framework.DependencyInjection.ServiceLookup
 {
     internal class ServiceScopeFactory : IServiceScopeFactory
     {
-        private static int _maxPooled = 32 * Environment.ProcessorCount;
+        private static int _capacity = 32 * Environment.ProcessorCount;
         private static readonly ConcurrentQueue<ServiceScope> _scopePool = new ConcurrentQueue<ServiceScope>();
 
         private readonly ServiceProvider _provider;
@@ -32,7 +32,7 @@ namespace Microsoft.Framework.DependencyInjection.ServiceLookup
         internal void PoolScope(ServiceScope scope)
         {
             // Benign race condition
-            if (_scopePool.Count < _maxPooled)
+            if (_scopePool.Count < _capacity)
             {
                 _scopePool.Enqueue(scope);
             }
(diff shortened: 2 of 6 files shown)


src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
```diff
@@ -19,12 +19,12 @@ internal class ServiceProvider : IServiceProvider, IDisposable
     {
         private readonly object _sync = new object();
 
-        private readonly Func<Type, Func<ServiceProvider, object>> _createServiceAccessor;
+        private Func<Type, Func<ServiceProvider, object>> _createServiceAccessor;
```
davidfowl: why this change?

benaadams: In error, has no state

src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
```diff
@@ -8,20 +8,37 @@ namespace Microsoft.Framework.DependencyInjection.ServiceLookup
     internal class ServiceScope : IServiceScope
     {
         private readonly ServiceProvider _scopedProvider;
+        private readonly ServiceScopeFactory _parentFactory;
 
         public ServiceScope(ServiceProvider scopedProvider)
         {
             _scopedProvider = scopedProvider;
         }
+        internal ServiceScope(ServiceProvider scopedProvider, ServiceScopeFactory parentFactory)
```
rynowak: Why internal for this? will the public constructor ever be used with these changes?

src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScopeFactory.cs
```diff
@@ -1,10 +1,16 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Collections.Concurrent;
+
 namespace Microsoft.Framework.DependencyInjection.ServiceLookup
 {
     internal class ServiceScopeFactory : IServiceScopeFactory
     {
+        private static int _maxPooled = 32 * Environment.ProcessorCount;
```
rynowak: `_capacity`?

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
```diff
@@ -145,15 +145,12 @@ internal IServiceCallSite GetResolveCallSite(IService service, ISet<Type> callSi
 
         public void Dispose()
         {
-            var disposables = Interlocked.Exchange(ref _disposables, null);
```
rynowak: This looks like it was done to avoid a double-dispose if `Dispose()` is called concurrently. With the current code it won't double-dispose any services but could call `Clear()` twice - is this a concern?

I suspect we're not too concerned about this scenario in practice. If there's any code you have operating on an SP while it's being disposed, you already have a bug.

src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
```diff
@@ -8,20 +8,37 @@ namespace Microsoft.Framework.DependencyInjection.ServiceLookup
...
+        }
 
         public IServiceProvider ServiceProvider
         {
             get { return _scopedProvider; }
         }
 
+        internal void Reset()
+        {
+            // Double wash
```
rynowak: Can you expand on this? what does this mean?

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
```diff
@@ -145,15 +145,12 @@ internal IServiceCallSite GetResolveCallSite(IService service, ISet<Type> callSi
...
+            IDisposable disposable;
+            while (_disposables.TryTake(out disposable))
             {
-                foreach (var disposable in disposables)
-                {
-                    disposable.Dispose();
-                }
+                disposable.Dispose();
             }
+            _resolvedServices.Clear();
```
rynowak: This needs a unit test - something where we can prove that that the following works correctly:

- Create a scoped SP
- Resolve a scoped service (instance 1)
- Dispose scoped SP
- Resolve a scoped service (instance 2)
- Assert that 1 and 2 aren't the same instance

src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
benaadams: Suppose not, missed it was an internal class

src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScopeFactory.cs
benaadams: Changed

src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
benaadams: Dispose again before reuse incase any extra disposables were added - overkill probably, but am concerned about information leakage between objects.

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
benaadams: It would just empty any extras; or do nothing if there were none. I have a little concern as its a concurrentbag which suggests items can be added in parallel; or out of order and with delays.

benaadams: In fact next comment is about the forced double dispose.

src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
benaadams: Phrased another way, original dispose is to drop any references so the can be GC'ed as the object is now kept; second Reset dispose is to make it safe for reuse.

I could swap ```_disposables``` to a backing object to make any adds error; then restore it in the Reset instead?

benaadams: Made this change

src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
```diff
@@ -19,9 +21,19 @@ public IServiceProvider ServiceProvider
             get { return _scopedProvider; }
         }
 
+        internal void Reset()
```
rynowak: Why internal here? It would also be helpful to capture your explanation of this in comments.

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
```diff
@@ -143,16 +144,28 @@ internal IServiceCallSite GetResolveCallSite(IService service, ISet<Type> callSi
             }
         }
 
+        internal void Reset()
+        {
+            if (_offlineDisposables == null)
+            {
+                throw new InvalidOperationException("ServiceProvider reused before being disposed");
```
rynowak: Do we have a resx in this project? Programmatic names like `ServiceProvider` need to be formatted in with parameters.

src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
benaadams: Not sure what to write... Its a method that is explicitly bonded to the pooling of ServiceScopeFactory so if the class were made public rather than internal for whatever reason the exposed api would be for it to function as a standalone class (which is why I left public constructor in); whereas the "internals" only work as a class pair.

Not hugely fussed on visibility though, can make public :)

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
benaadams: Done?

benaadams:
*Note:* still need to implement [unit test](https://github.com/aspnet/DependencyInjection/pull/289#discussion_r39432328) proposed by @rynowak, lost in *outdated diff* messages

src/Microsoft.Framework.DependencyInjection/ServiceLookup/ServiceScope.cs
davidfowl: I would leave it internal

benaadams:
See https://github.com/aspnet/HttpAbstractions/issues/402#issuecomment-140336990 for details on server run improvements (61.5k rps -> 1.49M rps)

rynowak:
:watch: - just waiting on a unit test for the scenario we discussed

benaadams:
Have the test, just trying to think of the right name XD

halter73:
:+1:

PR #291, opened 2015-09-16
[Design]: Changes to improve performance

davidfowl:
- Track disposables in a list with lock instead of a concurrent bag
- Allocate a single _createServiceAccessor instead of one per instance
- Removed syncObject and just locked the _resolvedServices

Measurements on the way...

/cc @rynowak @halter73 @pranavkm @lodejard @benaadams

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection/Internal/ConcurrentDictionaryExtensions.cs
+++ b/src/Microsoft.Framework.DependencyInjection/Internal/ConcurrentDictionaryExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics;
 
 namespace System.Collections.Concurrent
 {
@@ -8,20 +9,9 @@ namespace System.Collections.Concurrent
         // This lets us pass a state parameter allocation free GetOrAdd
         internal static TValue GetOrAdd<TKey, TValue, TArg>(this ConcurrentDictionary<TKey, TValue> dictionary, TKey key, Func<TKey, TArg, TValue> valueFactory, TArg arg)
         {
-            if (dictionary == null)
-            {
-                throw new ArgumentNullException(nameof(dictionary));
-            }
-
-            if (key == null)
-            {
-                throw new ArgumentNullException(nameof(key));
-            }
-
-            if (valueFactory == null)
-            {
-                throw new ArgumentNullException(nameof(valueFactory));
-            }
+            Debug.Assert(dictionary != null);
+            Debug.Assert(key != null);
+            Debug.Assert(valueFactory != null);
 
             while (true)
             {
--- a/src/Microsoft.Framework.DependencyInjection/Internal/ConcurrentDictionaryExtensions.cs
+++ b/src/Microsoft.Framework.DependencyInjection/Internal/ConcurrentDictionaryExtensions.cs
@@ -1,0 +1,43 @@
+﻿using System;
+
+namespace System.Collections.Concurrent
+{
+    internal static class ConcurrentDictionaryExtensions
+    {
+        // From https://github.com/dotnet/corefx/issues/394#issuecomment-69494764
+        // This lets us pass a state parameter allocation free GetOrAdd
+        internal static TValue GetOrAdd<TKey, TValue, TArg>(this ConcurrentDictionary<TKey, TValue> dictionary, TKey key, Func<TKey, TArg, TValue> valueFactory, TArg arg)
+        {
+            if (dictionary == null)
+            {
+                throw new ArgumentNullException(nameof(dictionary));
+            }
+
+            if (key == null)
+            {
+                throw new ArgumentNullException(nameof(key));
+            }
+
+            if (valueFactory == null)
+            {
+                throw new ArgumentNullException(nameof(valueFactory));
+            }
+
+            while (true)
+            {
+                TValue value;
+                if (dictionary.TryGetValue(key, out value))
+                {
+                    return value;
+                }
+
+                value = valueFactory(key, arg);
+                if (dictionary.TryAdd(key, value))
+                {
+                    return value;
+                }
+            }
+        }
+
+    }
+}
--- a/src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
+++ b/src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
@@ -17,14 +17,11 @@ namespace Microsoft.Framework.DependencyInjection
     /// </summary>
     internal class ServiceProvider : IServiceProvider, IDisposable
     {
-        private readonly object _sync = new object();
-
-        private readonly Func<Type, Func<ServiceProvider, object>> _createServiceAccessor;
         private readonly ServiceProvider _root;
         private readonly ServiceTable _table;
 
         private readonly Dictionary<IService, object> _resolvedServices = new Dictionary<IService, object>();
-        private ConcurrentBag<IDisposable> _disposables = new ConcurrentBag<IDisposable>();
+        private readonly List<IDisposable> _disposables = new List<IDisposable>();
 
         public ServiceProvider(IEnumerable<ServiceDescriptor> serviceDescriptors)
         {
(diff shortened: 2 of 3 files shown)


src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
```diff
@@ -145,25 +136,35 @@ internal IServiceCallSite GetResolveCallSite(IService service, ISet<Type> callSi
...
+
+                _disposables.Clear();
+                _resolvedServices.Clear();
             }
         }
 
         private object CaptureDisposable(object service)
         {
+            // Skip capturing disposables for the root
+            if (_root == this)
```
davidfowl: Nobody disposes the root scope, an argument can be made that it is possible but the interface for the top level (IServiceProvider) requires a downcast to IDisposable to do so and that's pretty rare unless you're scoped. We can let the GC handle those objects.

benaadams:
This halves allocations for DI

vs

src/Microsoft.Framework.DependencyInjection/Internal/ConcurrentDictionaryExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            while (true)
+            {
+                TValue value;
+                if (dictionary.TryGetValue(key, out value))
+                {
+                    return value;
+                }
+
+                value = valueFactory(key, arg);
```
rynowak: You can make this code a little clearer by getting rid of the `while` and changing the `TryAdd` to `GetOrAdd(key, value)`. If your add fails, it's because someone was contending with you to also add a value and so you'll get the value they added.

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
```diff
@@ -56,16 +47,16 @@ internal ServiceProvider(ServiceProvider parent)
         /// <returns></returns>
         public object GetService(Type serviceType)
         {
-            var realizedService = _table.RealizedServices.GetOrAdd(serviceType, _createServiceAccessor);
+            var realizedService = _table.RealizedServices.GetOrAdd(serviceType, (type, sp) => CreateServiceAccessor(type, sp), this);
```
rynowak: IMO worth having this in a static field and documenting why it is just so.

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
```diff
@@ -56,16 +47,16 @@ internal ServiceProvider(ServiceProvider parent)
...
             return realizedService.Invoke(this);
         }
 
-        private Func<ServiceProvider, object> CreateServiceAccessor(Type serviceType)
+        private static Func<ServiceProvider, object> CreateServiceAccessor(Type serviceType, ServiceProvider sp)
```
rynowak: `sp`? horrid

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
```diff
@@ -145,25 +136,35 @@ internal IServiceCallSite GetResolveCallSite(IService service, ISet<Type> callSi
...
+                return service;
+            }
+
             if (!object.ReferenceEquals(this, service))
             {
                 var disposable = service as IDisposable;
                 if (disposable != null)
                 {
-                    _disposables.Add(disposable);
+                    lock (_disposables)
```
rynowak: Do you care about the race condition that arises when someone is trying to add a disposable **while** the dispose methods are being run?

rynowak:
overall looks good - main point here seems to be assuming that the service provider is low-contention and we can do simpler/lighter things because of that.

src/Microsoft.Framework.DependencyInjection/Internal/ConcurrentDictionaryExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace System.Collections.Concurrent
+{
+    internal static class ConcurrentDictionaryExtensions
+    {
+        // From https://github.com/dotnet/corefx/issues/394#issuecomment-69494764
+        // This lets us pass a state parameter allocation free GetOrAdd
+        internal static TValue GetOrAdd<TKey, TValue, TArg>(this ConcurrentDictionary<TKey, TValue> dictionary, TKey key, Func<TKey, TArg, TValue> valueFactory, TArg arg)
+        {
+            if (dictionary == null)
```
pranavkm: If it's internal code, consider removing these exceptions and replacing it with `Debug.Assert` instead.

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
```diff
@@ -145,25 +136,35 @@ internal IServiceCallSite GetResolveCallSite(IService service, ISet<Type> callSi
 
         public void Dispose()
         {
-            var disposables = Interlocked.Exchange(ref _disposables, null);
-
-            if (disposables != null)
+            lock (_disposables)
```
pranavkm: Isn't the guideline to lock on a `_disposablesLock`?

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
```diff
@@ -312,7 +313,7 @@ private static Expression Lock(Expression providerExpression, Expression body)
             {
                 // The C# compiler would copy the lock object to guard against mutation.
                 // We don't, since we know the lock object is readonly.
-                var syncField = Expression.Field(providerExpression, "_sync");
+                var syncField = Expression.Field(providerExpression, "_resolvedServices");
```
pranavkm: Does `nameof(provider_resolvedServices)` work here?

src/Microsoft.Framework.DependencyInjection/Internal/ConcurrentDictionaryExtensions.cs
pranavkm: :+1:

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
rynowak: It's safe to lock on an arbitrary object when other code can't get at it. What you're suggesting is *safer* - but that's not the point. Your safety is not guaranteed.

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
davidfowl: Yea, that would be better

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
davidfowl: I started that way ya know :smile:

src/Microsoft.Framework.DependencyInjection/Internal/ConcurrentDictionaryExtensions.cs
davidfowl: Feels like this is easier to read.

rynowak: Follow your :heart: - but this gives me the impression that it will run an arbitrary number of times, but it will only ever run 1.5 times at most.

benaadams:
788k -> 384k (-51%)

If you didn't want to go full pooling, but were ok with partial pooling, you could pool the dictionary and list and clear them up in the disposable; which would trim it to 144k but limit the scope for the pooling.

788k -> 144k (-82%)

Though likely with that you'd need to return to an object for locking and add bool for dispose check

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
davidfowl: Not really, but we have no tests to verify it either. We don't even seem to check if we call GetService from a disposed service provider either.

davidfowl:
updated. 

Going to try a few more things:
- Don't allocate disposables until required.

src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
davidfowl: Don't have access to it.

davidfowl:
@benaadams New changes for ya.

pranavkm:
Looks :shipit:  to me. Will wait for @benaadams's numbers though.

benaadams:
Improved:

delayed-alloc

And down overall for the module/dll:

delayed-alloc

davidfowl:
Merged

PR #293, opened 2015-09-18
Prevent Dispose from being called twice on ServiceProvider

### LINKED ISSUE #292: ServiceProvider.Dispose() => Stack overflow
Introduced with bbb9acf. Not actually fixed by 2793fde.

Calls to https://github.com/aspnet/EntityFramework/blob/8667adaecfe8d96cd8929cf83939cae6fd43f189/src/EntityFramework.Core/Internal/DbContextServices.cs#L88
results in stack overflow.

Using Microsoft.Framework.DependencyInjection 1.0.0-beta8-15579

@davidfowl

natemcmaster:
Fix #292, bug introduced in bbb9acf101d4f9b3b818834c60572f67ec16ac63

### DIFF
--- a/src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
+++ b/src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
@@ -19,6 +19,7 @@ namespace Microsoft.Framework.DependencyInjection
     {
         private readonly ServiceProvider _root;
         private readonly ServiceTable _table;
+        private bool _disposeCalled;
 
         private readonly Dictionary<IService, object> _resolvedServices = new Dictionary<IService, object>();
         private List<IDisposable> _transientDisposables;
@@ -143,6 +144,13 @@ namespace Microsoft.Framework.DependencyInjection
         {
             lock (SyncObject)
             {
+                if (_disposeCalled)
+                {
+                    return;
+                }
+
+                _disposeCalled = true;
+
                 if (_transientDisposables != null)
                 {
                     foreach (var disposable in _transientDisposables)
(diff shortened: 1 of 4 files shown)


src/Microsoft.Framework.DependencyInjection/ServiceProvider.cs
```diff
@@ -211,6 +213,7 @@ private object GetEmptyIEnumerableOrNull(Type serviceType)
 
         private static MethodInfo MonitorEnterMethodInfo = GetMethodInfo<Action<object, bool>>((lockObj, lockTaken) => Monitor.Enter(lockObj, ref lockTaken));
         private static MethodInfo MonitorExitMethodInfo = GetMethodInfo<Action<object>>(lockObj => Monitor.Exit(lockObj));
+        private bool _disposeCalled;
```
davidfowl: move tis to the top of the class

test/Microsoft.Framework.DependencyInjection.Tests/ScopingContainerTestBase.cs
```diff
@@ -122,6 +122,16 @@ public void SelfResolveThenDispose()
         }
 
         [Fact]
+        public void SafelyDisposeNestedProviderReferences()
+        {
+            var container = CreateContainer();
```
davidfowl: Where is the assert?

natemcmaster: Implicit. This causes stack overflow without the fix.

davidfowl: Man more stackoverflows. Crazy :smile:

davidfowl:
:shipit:

PR #298, opened 2015-09-30
Downgrade to RTM packages for UWP

bricelam:


pranavkm:
What's the context for this change?

Eilon:
Do we have a mail thread on this? With the upcoming generations change I'm not sure this will be the right change and we might want to avoid changing this twice.

Eilon:
I started a mail thread.

Eilon:
:shipit:

PR #302, opened 2015-10-12
Sanitize loops

benaadams:
Esp. using jit recognised for (var i; i < a.length; i++) for bounds check elimination.

Replaces #299


pranavkm:
Thanks!

PR #304, opened 2015-10-15
Replace NotNullAttribute with thrown exceptions

pranavkm:


src/Microsoft.Extensions.DependencyInjection.Abstractions/ServiceDescriptor.cs
```diff
@@ -93,7 +123,8 @@ internal Type GetImplementationType()
                 return typeArguments[1];
             }
 
-            throw new ArgumentException(Resources.FormatNoImplementation(ServiceType));
+            Debug.Fail("ImplementationType, ImplementationInstance or ImplementationFactory must be non null");
```
pranavkm: We can't get here anymore due to the null checks in the ctor.

Eilon:
:shipit:

PR #309, opened 2015-10-23
Adding netcore50 TFM to support RTM UAP template

pranavkm:


pranavkm:
cc @natemcmaster

natemcmaster:
:shipit: (after travis checks out)

PR #312, opened 2015-10-26
Create a test specification package

pranavkm:

PR #313, opened 2015-10-29
Various tweaks

khellang:
- Rename: `container` -> `provider`
- Spelling: `sane` -> `same`
- Added asserts on transients resolved from scope

This targets the branch for PR #312

// @pranavkm


pranavkm:
Merged in https://github.com/aspnet/DependencyInjection/commit/b2d5aacdc8f438454c6374d189e7366471e942af

khellang:
@pranavkm When will this be pushed to NuGet? When RC1 hits the shelves? Or will it be on a CI feed somewhere?

pranavkm:
This is a RC2 item - so it won't get to nuget.org for a while. The packages should get on `https://www.myget.org/f/aspnetvnext` in some time (there's some churn in DNX around strong naming that's affecting package publish). aspnetvnext is equivalent to signed nightlies.

PR #318, opened 2015-11-04
Improve error message when dependency does not have a public constructor.

### LINKED ISSUE #310: Improve error message for "No parameterless constructor defined for this object"
This error message will manifest if a constructor is private, but is a dependency of another class. The following exception is thrown in beta7:

```
System.MissingMethodException: No parameterless constructor defined for this object.
```

It would be great if the message would identify the type with the missing constructor. Otherwise, one must resort to binary search debugging (commenting out various DI inputs) or similar efforts.

pranavkm:
Fixes #310

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection/Resources.resx
+++ b/src/Microsoft.Extensions.DependencyInjection/Resources.resx
@@ -137,4 +137,8 @@
   <data name="TypeCannotBeActivated" xml:space="preserve">
     <value>Cannot instantiate implementation type '{0}' for service type '{1}'.</value>
   </data>
+  <data name="NoConstructorMatch" xml:space="preserve">
+    <value>Unable to locate suitable constructor for type '{0}'. Ensure the type is concrete and services are registered for all parameters of a public constructor.</value>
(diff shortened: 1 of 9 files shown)


pranavkm:
cc @Eilon

src/Microsoft.Extensions.DependencyInjection/Resources.resx
```diff
@@ -137,4 +137,8 @@
   <data name="TypeCannotBeActivated" xml:space="preserve">
     <value>Cannot instantiate implementation type '{0}' for service type '{1}'.</value>
   </data>
+  <data name="NoConstructorMatch" xml:space="preserve">
+    <value>Unable to locate suitable constructor for type '{0}'. Ensure the type is concrete and services are registered for all parameters of a public constructor.</value>
```
Eilon: Change first sentence to: `A suitable constructor for type '{0}' couldn't be located.`

(All I did was translate from Robot-ese to Human-ese :smile: )

Eilon:
:shipit:

PR #321, opened 2015-11-10
Making types in Microsoft.Extensions.DependencyInjection public

### LINKED ISSUE #315: Consider making types in Microsoft.Extensions.DependencyInjection public
For instance:
https://github.com/aspnet/DependencyInjection/blob/dev/src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs#L18

We could place them in an `Internal` namespace if necessary..

pranavkm:
Fixes #315

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection/Internal/ConcurrentDictionaryExtensions.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/Internal/ConcurrentDictionaryExtensions.cs
@@ -1,4 +1,6 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System.Diagnostics;
 
 namespace System.Collections.Concurrent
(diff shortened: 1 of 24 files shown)


Eilon:
And what's the real reason for this? What does this help?

pranavkm:
Nothing in particular?

Eilon:
Then let's not do it now? :smile: These are some pretty obscure types...

pranavkm:
:crying_cat_face:

Eilon:
I know, I know, so sad :sob:

DovydasNavickas:
Hmmm... I found this while looking at a way to use DI for testing.
Of course, there is `TestServer` with the whole `Startup.cs` and its `ConfigureServices` and `Configure` mechanism, but my application does polling from service bus and not HTTP requests handling, which makes it quite hard to use the TestServer.
Http messages only control the polling and processing mechanisms and report current status, if requested.
I suppose there is no way to get the same DI implementation that I get in web application, is there?
Maybe I'm doing something fundamentally wrong?

pranavkm:
@DovydasNavickas could you file a new issue stating what you're trying to do and why you can't get it to work?

DovydasNavickas:
Done: https://github.com/aspnet/DependencyInjection/issues/445

PR #324, opened 2015-11-12
Rename AddInstance to AddSingleton

### LINKED ISSUE #319: Rename AddInstance to AddSingleton
The name `AddInstance` doesn't make it super clear what the lifetime is of the object. (If you stop and think about it, it can only be Singleton, but every time I call that method I have to stop and think about it.)

So, we should rename `AddInstance` to `AddSingleton`, or perhaps `AddSingletonInstance`.

pranavkm:
Fixes #319


pranavkm:
cc @Eilon \ @davidfowl

davidfowl:
@pranavkm Everything will start to fail because we have warnings as errors. Make sure you do a pass through the other repos.

:shipit:

pranavkm:
@Eilon had some concerns about using `ObsoleteAttribute`.

Eilon:
Yeah I'm not sure why we'd go out of the way to mark APIs as obsolete. There are much larger changes than this that have gone through. That's why we have the Announcements repo.

pranavkm:
Ok. I'll get rid of `AddInstance`

PR #330, opened 2015-11-20
Move ServiceContainer specific tests out of specifications.

### LINKED ISSUE #325: Remove AttemptingToResolveNonexistentServiceIndirectlyThrows test from test specification
We have a test in the specification that verifies that the service provider throws if a dependent service cannot be located. This is a behavior of the service container and we should move it out and make it part of `ServiceProviderContainerTests`

pranavkm:
Fixes #325


pranavkm:
cc @Eilon

Eilon:
:shipit:

Issue #340, opened 2015-12-18
Allow services to decorate each other by preventing recursive resolution


tuespetre:
### User Story / Use Case Scenario

As a developer, I want to extend a previously registered service by wrapping it with another service that exposes the same interface.

See: https://github.com/aspnet/Mvc/issues/3739
### Example interface and implementations

```
public interface IMyService
{
    void DoSomething();
}

public partial class DefaultMyService
{
    public partial void DoSomething();
}

public partial class DecoratorForMyService
{
    private IMyService decorated;

    public DecoratorForMyService(IMyService decorated)
    {
        this.decorated = decorated;
    }

    public partial void DoSomething();
}
```
#### Using constructor resolution

```
var collection = new ServiceCollection();
collection.AddTransient<IMyService, DefaultMyService>();
collection.AddTransient<IMyService, DecoratorForMyService>();
var provider = collection.BuildServiceProvider();
var service = collection.GetService<IMyService>(); // Returns the constructed DecoratorForMyService
```
#### Using implementation factories

```
var collection = new ServiceCollection();
collection.AddTransient<IMyService, DefaultMyService>();
collection.AddTransient<IMyService>(_provider =>
{
    // Returns the previously registered IMyService.
    // Calling _provider.GetService<IEnumerable<IMyService>>() here
    // would return a collection containing all registered IMyServices
    // except for this one.
    var decorated = _provider.GetService<IMyService>(); 
    return new DecoratorForMyService(decorated);
});
var provider = collection.BuildServiceProvider();
var service = collection.GetService<IMyService>(); // Returns the constructed DecoratorForMyService
```
### Behavioral Considerations

Currently, the above two approaches result in 'Circular Dependency' exception and a `StackOverflowException`, respectively. An implementation that allowed for the above scenarios would effectively prevent the `StackOverflowException` entirely, and ~~throw the 'Circular Dependency' exception only when there are no other previously registered services for the type that is being resolved~~ result in circular dependency scenarios throwing the "Unable to resolve service..." kind of exception, which should in itself sufficiently convey the circularity for those cases.

davidfowl:
We're not going to support something like this. The problem is it's a very specific feature and we'd need other DI systems to work this way. Or we could just give up and force you to use ours

davidfowl:
For research can you take a look at some other DI systems and see how they solve those problem. If they even allow it

tuespetre:
[...]

PR #341, opened 2015-12-18
Address #340

tuespetre:
- Correct 'ambigious' typo (should be 'ambiguous')
- Instead of circular dependency detection for constructors or
  StackOverFlowExceptions for factories, pull out each currently resolving
  service from the ServiceEntry. This allows factories and constructors
  to access the other services registered under the same type
- Part of implementing this also requires that service resolution
  callsites be cached by the actual service, rather than just the type


Eilon:
https://github.com/aspnet/DependencyInjection/issues/340 was moved to the backlog, so closing this for now.

PR #353, opened 2016-01-29
Make TryAdd* methods return bool

khellang:
It's nice if the methods indicate whether something was added or not :smile:

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection.Abstractions/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.DependencyInjection.Abstractions/Extensions/ServiceCollectionExtensions.cs
@@ -130,7 +130,7 @@ namespace Microsoft.Extensions.DependencyInjection.Extensions
 
             var descriptor = ServiceDescriptor.Transient(service, service);
 
-            return TryAdd(collection, descriptor);
+            return collection.TryAdd(descriptor);
         }
 
         public static bool TryAddTransient(
@@ -155,7 +155,7 @@ namespace Microsoft.Extensions.DependencyInjection.Extensions
 
             var descriptor = ServiceDescriptor.Transient(service, implementationType);
 
-            return TryAdd(collection, descriptor);
+            return collection.TryAdd(descriptor);
         }
 
         public static bool TryAddTransient(
@@ -180,7 +180,7 @@ namespace Microsoft.Extensions.DependencyInjection.Extensions
 
             var descriptor = ServiceDescriptor.Transient(service, implementationFactory);
 
-            return TryAdd(collection, descriptor);
+            return collection.TryAdd(descriptor);
         }
 
         public static bool TryAddTransient<TService>(this IServiceCollection collection)
@@ -191,7 +191,7 @@ namespace Microsoft.Extensions.DependencyInjection.Extensions
                 throw new ArgumentNullException(nameof(collection));
             }
 
-            return TryAddTransient(collection, typeof(TService), typeof(TService));
+            return collection.TryAddTransient(typeof(TService), typeof(TService));
         }
 
         public static bool TryAddTransient<TService, TImplementation>(this IServiceCollection collection)
@@ -203,7 +203,7 @@ namespace Microsoft.Extensions.DependencyInjection.Extensions
                 throw new ArgumentNullException(nameof(collection));
             }
 
-            return TryAddTransient(collection, typeof(TService), typeof(TImplementation));
+            return collection.TryAddTransient(typeof(TService), typeof(TImplementation));
         }
 
         public static bool TryAddTransient<TService>(
@@ -230,7 +230,7 @@ namespace Microsoft.Extensions.DependencyInjection.Extensions
 
             var descriptor = ServiceDescriptor.Scoped(service, service);
 
-            return TryAdd(collection, descriptor);
+            return collection.TryAdd(descriptor);
         }
 
         public static bool TryAddScoped(
@@ -255,7 +255,7 @@ namespace Microsoft.Extensions.DependencyInjection.Extensions
 
             var descriptor = ServiceDescriptor.Scoped(service, implementationType);
 
-            return TryAdd(collection, descriptor);
+            return collection.TryAdd(descriptor);
         }
 
         public static bool TryAddScoped(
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.DependencyInjection.Abstractions/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -108,13 +111,10 @@ public static class ServiceCollectionExtensions
                 throw new ArgumentNullException(nameof(descriptors));
             }
 
-            foreach (var d in descriptors)
-            {
-                collection.TryAdd(d);
-            }
+            return descriptors.All(collection.TryAdd);
```
khellang: Not sure what to return here...

src/Microsoft.Extensions.DependencyInjection.Abstractions/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -504,10 +517,7 @@ public static void TryAddSingleton<TService>(this IServiceCollection collection,
                 throw new ArgumentNullException(nameof(descriptors));
             }
 
-            foreach (var d in descriptors)
-            {
-                services.TryAddEnumerable(d);
-            }
+            return descriptors.All(services.TryAdd);
```
khellang: Same here... Is this expected?

henkmollema:
Shouldn't this be `TryAddEnumerable`?

khellang:
@henkmollema Shouldn't what be?

henkmollema:
`return descriptors.All(services.TryAdd)`

src/Microsoft.Extensions.DependencyInjection.Abstractions/Extensions/ServiceCollectionExtensions.cs
henkmollema: I meant this line.

smbecker:
They used to return bool but were changed to return void as part of cbd74e57f28da3f4fc90174b4cf91042fa07a323. I'm not sure why. Granted, I happen to like having the bool return type.

src/Microsoft.Extensions.DependencyInjection.Abstractions/Extensions/ServiceCollectionExtensions.cs
khellang: This method is already called `TryAddEnumerable` :smile:

pranavkm:
cc @Eilon for API changes.

Eilon:
Hmm the behavior of the collection APIs is concerning to me. What is the scenario for this change? That would likely help us figure out the desired behavior.

src/Microsoft.Extensions.DependencyInjection.Abstractions/Extensions/ServiceCollectionExtensions.cs
henkmollema: You call `TryAdd` now in stead of `TryAddEnumerable` now.

khellang:
> What is the scenario for this change? That would likely help us figure out the desired behavior.

You mean for the collection APIs? There is no scenario - at least not from my perspective. It was just for consistency.

I agree that it doesn't quite make sense, whether you return `true` for all or any registrations. That's why I said

> Not sure what to return here...

khellang:
However, back in https://github.com/aspnet/DependencyInjection/commit/cbd74e57f28da3f4fc90174b4cf91042fa07a323#diff-7363faa647a95e5a8e6f6696d5d3d1bfL74, it used to return the equivalent of a LINQ `.Any`...

Eilon:
Hmm so is there a scenario for any of the `bool` return values? What would a caller do if it was true/false? At this point in the product cycle we're looking for changes that have significant impact and I'm not yet seeing how this fits.

khellang:
> Hmm so is there a scenario for any of the bool return values?

Yes. For many reasons, it would be nice to know the `TryAdd*` methods *actually* added anything to the collection when you call it.
I'd argue that there should be a convenience method to check if the collection already contains registrations for a given type as well.

Eilon:
Heh, well, I'm curious what are the "many reasons" :smile:

khellang:
Never mind.

Eilon:
@khellang honestly I can't see the use for it - it just hasn't come up in years. But I'd like to know more about what I'm missing here.

khellang:
I honestly don't get why this matters. It's a case where you have some information that can be valuable to someone and have the possibility to give it to them, why not do it? If someone asks for it and takes the time to PR it, there obviously has to be some scenarios for it.

> it just hasn't come up in years

Have you ever thought that you as framework devs might have different use cases than application users? This is still in RC, and has barely reached "mainstream" devs yet, and probably won't in some time. Trust me - this'll come up time and time again :wink:

Eilon:
@khellang there was a problem in the PR that wasn't resolved - what should the behavior be for the collection-based methods. Understanding the scenario in greater detail would help determine the right design.

For example, if a collection-based method returns `false`, what should the caller do? Try each service individually to identify the error? (Which isn't even possible anymore because all future attempts would always return `false`.) And if we can't come up with the right behavior for the collection-based methods, of what use are they with respect to the `bool` success behavior?

Ultimately I think the fact that it hasn't come up in years is a good indication that it's really not very important. I'm not denying that it's never important to anybody, but that's different from being broadly important.

khellang:
> Ultimately I think the fact that it hasn't come up in years is a good indication that it's really not very important.

Where does all these years come from? The framework isn't that old. I think the "the reason it hasn't come up in years" is probably because you, Microsoft, the framework devs, don't have many scenarios where it's important (even though it was originally written this way, for "some reason"). There aren't many other users of this yet.

The reason I PR'd this in the first place was because *several* people have asked for this in the AspNetCore JabbR room over time and @davidfowl asked me to. I think more people will want this as time goes by. I don't think there's a single container in the .NET space that doesn't allow you to check for an existing registration before adding one.

I don't really care about the enumerable overloads. As I mentioned, those were just changed for consistency.

The thing I really don't understand is; if you look away from the enumerable overloads, what's so horrible with providing the users with some information they might need/want. Are you compromising anything else? Please help me understand.

Eilon:
@khellang I think it's reasonable to understand the scenarios for a change before committing to the change. So far I haven't been able to understand why the return value is needed.

Re: Years - well yeah, this assembly has been around for 2+ years, and there hasn't been much, if any, demand for this change. So I'd like to understand what has changed.

Worthaboutapig:
Well, one good reason is consistency- it's generally accepted that TryX returns success or failure (e.g. int.TryParse).  And if enumerables doesn't fit, then perhaps it's not an appropriate use, or an option might be returning false on any failure and 'out'ing a list of the failed registrations.

PR #354, opened 2016-01-29
Adds generic factory overloads with both TService and TImplementation for TryAdd*

khellang:
This sits on top of #353

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection.Abstractions/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.DependencyInjection.Abstractions/Extensions/ServiceCollectionExtensions.cs
@@ -130,7 +130,7 @@ namespace Microsoft.Extensions.DependencyInjection.Extensions
 
             var descriptor = ServiceDescriptor.Transient(service, service);
 
-            return TryAdd(collection, descriptor);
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.DependencyInjection.Abstractions/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -504,10 +544,7 @@ public static void TryAddSingleton<TService>(this IServiceCollection collection,
                 throw new ArgumentNullException(nameof(descriptors));
             }
 
-            foreach (var d in descriptors)
-            {
-                services.TryAddEnumerable(d);
-            }
+            return descriptors.All(services.TryAdd);
```
davidfowl: I'm not sure this is equivalent.

khellang: Oh, good catch!

henkmollema: That's what I meant in the other PR

khellang:
I guess someone else can PR this if they want it later.

PR #362, opened 2016-02-17
[Fixes aspnet/common#88] Make ActivatorUtilities depend on Common implementation

javiercn:


davidfowl:
Is it the same now?

src/Microsoft.Extensions.DependencyInjection.Abstractions/project.json
```diff
@@ -1,26 +1,32 @@
...
-        "System.Globalization": "4.0.11-*",
-        "System.Linq": "4.0.2-*",
-        "System.Linq.Expressions": "4.0.11-*",
-        "System.Reflection": "4.1.0-*",
-        "System.Resources.ResourceManager": "4.0.1-*"
-      }
+    "version": "1.0.0-*",
+    "description": "Contracts for dependency injection.",
+    "repository": {
+        "type": "git",
```
pranavkm: 2-space indentation for json files.

pranavkm:
:shipit:

javiercn:
@davidfowl Kind of, we have a sources package in Common and we depend on that package on DI. DI also defines this class as a public class and acts as a facade for the sources implementation. Everywhere in our repositories we use the DI version, the only place in which we use the version in Common is in HttpAbstraction (Or in the future, in any case where we need this functionality and we can't take an dependency on DI). Using the version in Common is discouraged and this instance should be used instead.

PR #369, opened 2016-03-02
Switching to minimal TFMs

pranavkm:


pranavkm:
cc @davidfowl

davidfowl:
:shipit: I need to investigate why you needed to add System.Runtime.

khellang:
:heart:

src/Microsoft.Extensions.DependencyInjection.Abstractions/ServiceDescriptor.cs
```diff
@@ -123,7 +123,7 @@ internal Type GetImplementationType()
                 return typeArguments[1];
             }
 
-            Debug.Fail("ImplementationType, ImplementationInstance or ImplementationFactory must be non null");
+            Debug.Assert(false, "ImplementationType, ImplementationInstance or ImplementationFactory must be non null");
```
halter73: What platform doesn't have Debug.Fail btw?

pranavkm: netstandard1.0 which is the only TFM we're targeting in this package.

PR #371, opened 2016-03-05
Simplify codegen of scoped and singleton expressions

davidfowl:
/cc @halter73 @pranavkm

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
@@ -208,12 +208,22 @@ namespace Microsoft.Extensions.DependencyInjection
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
```diff
@@ -279,16 +289,7 @@ public ScopedCallSite(IService key, IServiceCallSite serviceCallSite)
...
-                lock (provider._resolvedServices)
-                {
-                    if (!provider._resolvedServices.TryGetValue(_key, out resolved))
-                    {
-                        resolved = _serviceCallSite.Invoke(provider);
-                        provider._resolvedServices.Add(_key, resolved);
-                    }
-                }
-                return resolved;
+                return provider.GetOrAdd(_key, () => _serviceCallSite.Invoke(provider));
```
halter73: Does this create a closure that didn't exist before? Does this fix a bug or measurably improve perf?

PR #377, opened 2016-03-08
Simplify codegen for scoped and singleton services

davidfowl:


davidfowl:
Ok turns out this is in general less efficient because of the magic the C# compiler does. It's fixable but it requires more codegen:

Consider:

```C#
provider.CaptureDisposable(
     new A(
        provider.GetOrAdd(key, p => 
            new B(p._root.GetOrAdd(key, new C())))
```

The compiler will hoist the lambdas into a static method:

```

public static Func<ServiceProvider, object> ACallback;
public static Func<ServiceProvider, object> BCallback;

{
    if (BCallback != null)
    {
        BCallback = p => new C();
    }

    if (ACallback != null)
    {
        ACallback = p => new B(p._root.GetOrAdd(key, BCallback));
    }

    provider.CaptureDisposable(new A(provider.GetOrAdd(key, ACallback));
}
```

davidfowl:
/cc @pranavkm @halter73 

Tapping out :smile:

PR #381, opened 2016-03-10
Don't reference facades in NuSpec

bricelam:
These can be removed entirely after dotnet/cli#164


pranavkm:
:shipit:

bricelam:
Not needed after 555a0ec91477cbb5c33a158a5d3bfe912edd6888

PR #383, opened 2016-03-11
Use a non primtive type

### LINKED ISSUE #382: Should it be possible to add primitive types?
Hi! [There's a unit test](https://github.com/aspnet/DependencyInjection/blob/dev/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/DependencyInjectionSpecificationTests.cs#L504) that adds a primitive type as a singleton to the collection: 
```csharp
collection.AddSingleton("Hello World");
```
Should this really be possible? What would be a usage scenario for this? 

Castle Windsor for example [doesn't support it](https://github.com/castleproject/Windsor/blob/master/src/Castle.Windsor/Core/ComponentModel.cs#L376).

pranavkm:
Fixes #382


pranavkm:
cc @cwe1ss \ @davidfowl

davidfowl:
:shipit:

cwe1ss:
thx guys!

PR #384, opened 2016-03-11
Enable 3rd party DI containers to handle requests for required services.

kichalla:
[Fixes ##378] Add interface to allow container implementors to expose a more native service lookup for required services

@davidfowl

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection.Abstractions/ISupportRequiredService.cs
+++ b/src/Microsoft.Extensions.DependencyInjection.Abstractions/ISupportRequiredService.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// Optional contract used by <see cref="ServiceProviderExtensions.GetRequiredService{T}(IServiceProvider)"/>
+    /// to resolve services if supported by <see cref="IServiceProvider"/>.
+    /// </summary>
+    public interface ISupportRequiredService
+    {
+        /// <summary>
+        /// Gets service of type <paramref name="serviceType"/> from the <see cref="IServiceProvider"/> implementing
+        /// this interface.
+        /// </summary>
+        /// <param name="serviceType">An object that specifies the type of service object to get.</param>
+        /// <returns>A service object of type <paramref name="serviceType"/>.
+        /// Throws an exception if the <see cref="IServiceProvider"/> cannot create the object.</returns>
+        object GetRequiredService(Type serviceType);
+    }
+}
--- a/src/Microsoft.Extensions.DependencyInjection.Abstractions/ServiceProviderExtensions.cs
+++ b/src/Microsoft.Extensions.DependencyInjection.Abstractions/ServiceProviderExtensions.cs
@@ -47,8 +47,13 @@ namespace Microsoft.Extensions.DependencyInjection
                 throw new ArgumentNullException(nameof(serviceType));
             }
 
-            var service = provider.GetService(serviceType);
+            var requiredServiceSupportingProvider = provider as ISupportRequiredService;
+            if (requiredServiceSupportingProvider != null)
+            {
+                return requiredServiceSupportingProvider.GetRequiredService(serviceType);
+            }
 
+            var service = provider.GetService(serviceType);
             if (service == null)
             {
                 throw new InvalidOperationException(Resources.FormatNoServiceRegistered(serviceType));
--- a/test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
+++ b/test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
@@ -25,6 +25,19 @@ namespace Microsoft.Extensions.DependencyInjection
         }
 
         [Fact]
+        public void ISupportRequiredService_GetRequiredService_Returns_CorrectService()
+        {
+            // Arrange
+            var serviceProvider = new RequiredServiceSupportingProvider();
+
+            // Act
+            var service = serviceProvider.GetRequiredService<IBar>();
+
+            // Assert
+            Assert.IsType<Bar1>(service);
+        }
+
+        [Fact]
         public void GetRequiredService_Throws_WhenNoServiceRegistered()
         {
             // Arrange
@@ -36,6 +49,16 @@ namespace Microsoft.Extensions.DependencyInjection
         }
 
         [Fact]
+        public void ISupportRequiredService_GetRequiredService_Throws_WhenNoServiceRegistered()
+        {
+            // Arrange
+            var serviceProvider = new RequiredServiceSupportingProvider();
+
+            // Act + Assert
+            ExceptionAssert.Throws<RankException>(() => serviceProvider.GetRequiredService<IFoo>());
+        }
+
+        [Fact]
         public void NonGeneric_GetRequiredService_Throws_WhenNoServiceRegistered()
         {
             // Arrange
@@ -44,6 +67,16 @@ namespace Microsoft.Extensions.DependencyInjection
             // Act + Assert
             ExceptionAssert.Throws<InvalidOperationException>(() => serviceProvider.GetRequiredService(typeof(IFoo)),
                 $"No service for type '{typeof(IFoo)}' has been registered.");
+        }
+
+        [Fact]
+        public void ISupportRequiredService_NonGeneric_GetRequiredService_Throws_WhenNoServiceRegistered()
+        {
+            // Arrange
+            var serviceProvider = new RequiredServiceSupportingProvider();
+
+            // Act + Assert
+            ExceptionAssert.Throws<RankException>(() => serviceProvider.GetRequiredService(typeof(IFoo)));
         }
 
         [Fact]
@@ -219,5 +252,23 @@ namespace Microsoft.Extensions.DependencyInjection
         public class Bar1 : IBar { }
 
         public class Bar2 : IBar { }
+
+        private class RequiredServiceSupportingProvider : IServiceProvider, ISupportRequiredService
+        {
+            object ISupportRequiredService.GetRequiredService(Type serviceType)
+            {
+                if (serviceType == typeof(IBar))
+                {
+                    return new Bar1();
+                }
+
+                throw new RankException();
+            }
+
+            object IServiceProvider.GetService(Type serviceType)
+            {
+                throw new NotSupportedException();
+            }
+        }
     }
 }


test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -219,5 +241,20 @@ public interface IBar { }
...
+            public const string ErrorMessageFormat = "Required service {0} was not found.";
+
+            public object GetRequiredService(Type serviceType)
+            {
+                throw new InvalidOperationException(string.Format(ErrorMessageFormat, serviceType.Name));
+            }
+
+            public object GetService(Type serviceType)
+            {
+                return false;
```
dadhi: That is strange returned result. I assume it is only for test?

src/Microsoft.Extensions.DependencyInjection.Abstractions/IRequiredServiceAware.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// Implementation which gets invoked on the <see cref="IServiceProvider"/> when
```
pranavkm: > Optional contract used by `ServiceProviderExtensions.GetRequiredService(IServiceProvider, Type)` to resolve services if supported by `IServiceProvider`.

or something like that

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -36,6 +36,17 @@ public void GetRequiredService_Throws_WhenNoServiceRegistered()
         }
 
         [Fact]
+        public void IRequiredServiceAwareService_GetRequiredService_Throws_WhenNoServiceRegistered()
+        {
+            // Arrange
+            var serviceProvider = new RequiredServiceAwareServiceProvider();
+
+            // Act + Assert
+            ExceptionAssert.Throws<InvalidOperationException>(() => serviceProvider.GetRequiredService<IFoo>(),
```
pranavkm: Could you throw a different exception here? It's hard to tell if the exception comes from your service provider \ from the extension method.

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -219,5 +241,20 @@ public interface IBar { }
         public class Bar1 : IBar { }
 
         public class Bar2 : IBar { }
+
+        private class RequiredServiceAwareServiceProvider : IServiceProvider, IRequiredServiceAware
+        {
+            public const string ErrorMessageFormat = "Required service {0} was not found.";
```
pranavkm: Use a different exception type - so you don't have to match strings

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
pranavkm: You could `throw new NotSupportedException` here since it's never going to get called.

pranavkm:
:watch:

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
kichalla: :+1:

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
kichalla: :+1:

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -47,6 +58,17 @@ public void NonGeneric_GetRequiredService_Throws_WhenNoServiceRegistered()
         }
 
         [Fact]
+        public void IRequiredServiceAware_NonGeneric_GetRequiredService_Throws_WhenNoServiceRegistered()
+        {
+            // Arrange
+            var serviceProvider = new RequiredServiceAwareServiceProvider();
+
+            // Act + Assert
+            ExceptionAssert.Throws<InvalidOperationException>(() => serviceProvider.GetRequiredService(typeof(IFoo)),
```
smbecker: Wouldn't this just hit the RequiredServiceAwareServiceProvider.GetRequiredService method directly rather than the extension method given than it is using type-inference and the class exposes GetRequiredService as a public method?

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -219,5 +241,20 @@ public interface IBar { }
         public class Bar1 : IBar { }
 
         public class Bar2 : IBar { }
+
+        private class RequiredServiceAwareServiceProvider : IServiceProvider, IRequiredServiceAware
```
smbecker: Perhaps these methods should be implemented explicitly?

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
pranavkm: Yup. Making it an explicit implementation would avoid accidentally referencing the method rather than the extension.

src/Microsoft.Extensions.DependencyInjection.Abstractions/IRequiredServiceAware.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// <see cref="ServiceProviderExtensions.GetRequiredService{T}(IServiceProvider)"/> or
+    /// <see cref="ServiceProviderExtensions.GetRequiredService(IServiceProvider, Type)"/> is called.
+    /// </summary>
+    public interface IRequiredServiceAware
```
HaoK: How about ISupportRequiredService :)

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
kichalla: Good catch.

src/Microsoft.Extensions.DependencyInjection.Abstractions/IRequiredServiceAware.cs
kichalla: I like it :) @davidfowl ?

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -36,6 +36,18 @@ public void GetRequiredService_Throws_WhenNoServiceRegistered()
...
 
         [Fact]
+        public void IRequiredServiceAwareService_GetRequiredService_Throws_WhenNoServiceRegistered()
+        {
+            // Arrange
+            var serviceProvider = new RequiredServiceAwareServiceProvider();
+
+            // Act + Assert
+            ExceptionAssert.Throws<InvalidOperationException>(
+                () => ((IServiceProvider)serviceProvider).GetRequiredService<IFoo>(),
```
kichalla: @smbecker @pranavkm How does this look like? (yeah I could avoid `var` here and avoided the cast, but we usually are using `var` everywhere..)

pranavkm: Just make the method explicit.

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -219,5 +243,20 @@ public interface IBar { }
         public class Bar1 : IBar { }
 
         public class Bar2 : IBar { }
+
+        private class RequiredServiceAwareServiceProvider : IServiceProvider, IRequiredServiceAware
+        {
+            public const string ErrorMessageFormat = "Boooo!! Required service {0} was not found";
+
+            public object GetRequiredService(Type serviceType)
```
pranavkm: `object IRequireServiceAware.GetRequiredService(Type serviceType)`

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -36,6 +36,18 @@ public void GetRequiredService_Throws_WhenNoServiceRegistered()
         }
 
         [Fact]
+        public void IRequiredServiceAwareService_GetRequiredService_Throws_WhenNoServiceRegistered()
```
pranavkm: Do we need additional tests where `IRequiredService` succeeds?

kichalla: Added a test now.

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -219,5 +256,25 @@ public interface IBar { }
...
+            public const string ErrorMessageFormat = "Boooo!! Required service {0} was not found";
+
+            object IRequiredServiceAware.GetRequiredService(Type serviceType)
+            {
+                if (serviceType == typeof(IBar))
+                {
+                    return new Bar1();
+                }
+
+                throw new InvalidOperationException(string.Format(ErrorMessageFormat, serviceType.Name));
```
pranavkm: Much easier to just return a custom exception type. `throw new RankException` or some thing like that and you don't have to verify the exception message.

src/Microsoft.Extensions.DependencyInjection.Abstractions/IRequiredServiceAware.cs
khellang: Or just `IRequiredServiceProvider`? :wink:

davidfowl: `ISupportRequiredService` sounds better as it describes the capability

khellang: And `IRequiredServiceProvider` doesn't? I'm really not a fan of the `IProvideServices`, `ICreateLoggers`, `IActivateControllers` naming, compared to `IServiceProvider`, `ILoggerFactory` and `IControllerActivator`. It would be nice if it were consistent. Just my 2¢ :smile:

davidfowl: One difference here is this is a capability, so it's better to look at other optional interfaces to compare apples to apples. This is more like our feature interfaces than it is like the other ones you mention. `IServiceProvider` and `IServiceScopeFactory` are services in the container itself, I'm not sure this should follow that pattern. This is more of an opportunistic downcast to enable `IServiceProvider` implementors to also provide a GetRequiredService method

davidfowl: #naming is hard

kichalla:
:up: :date: @pranavkm @davidfowl

davidfowl:
:shipit:

src/Microsoft.Extensions.DependencyInjection.Abstractions/ISupportRequiredService.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// Optional contract used by <see cref="ServiceProviderExtensions.GetRequiredService{T}(IServiceProvider)"/>
+    /// to resolve services if supported by IServiceProvider.
```
pranavkm: > by `<see cref="IServiceProvider"/>`.

src/Microsoft.Extensions.DependencyInjection.Abstractions/ServiceProviderExtensions.cs
```diff
@@ -47,8 +47,13 @@ public static object GetRequiredService(this IServiceProvider provider, Type ser
                 throw new ArgumentNullException(nameof(serviceType));
             }
 
-            var service = provider.GetService(serviceType);
+            var requiredServiceSupportingProvider = provider as ISupportRequiredService ;
```
pranavkm: Extra space before `;`

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderExtensionsTest.cs
```diff
@@ -36,6 +49,16 @@ public void GetRequiredService_Throws_WhenNoServiceRegistered()
         }
 
         [Fact]
+        public void ISupportRequiredService_GetRequiredService_Throws_WhenNoServiceRegistered()
+        {
+            // Arrange
+            var serviceProvider = new RequiredServiceSupportingProvider();
+
+            // Act + Assert
+            ExceptionAssert.Throws<RankException>(() => serviceProvider.GetRequiredService<IFoo>());
```
pranavkm: :100:

pranavkm:
:shipit:

PR #387, opened 2016-03-17
Build with xml-docs-test only if src folder exists.

cesarblum:
cc @victorhurdugaci


victorhurdugaci:
I don't think this target is needed anymore. Maybe a quick workaround is to remove the body of the sake target

victorhurdugaci:
cc @pranavkm

victorhurdugaci:
This is not needed anymore

pranavkm:
Yup. The new CLI takes care of this.

PR #399, opened 2016-04-09
Cross-compile to netcore50 to support EF on UWP

natemcmaster:
Part of https://github.com/aspnet/EntityFramework/pull/5021

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection/project.json
+++ b/src/Microsoft.Extensions.DependencyInjection/project.json
@@ -28,9 +28,6 @@
         "System.Runtime.Extensions": "4.1.0-*",
         "System.Threading": "4.0.11-*",
         "System.Threading.Tasks": "4.0.11-*"
-      },
-      "runtimes": {
-        "any": {}
       }
     },
     "netcore50": {
@@ -44,14 +41,6 @@
         "System.Runtime.Extensions": "4.0.10",
         "System.Threading": "4.0.10",
         "System.Threading.Tasks": "4.0.10"
-      },
-      "runtimes": {
-        "win10-arm": {},
-        "win10-arm-aot": {},
-        "win10-x86": {},
-        "win10-x86-aot": {},
-        "win10-x64": {},
-        "win10-x64-aot": {}
       }
     }
   }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.DependencyInjection/project.json
```diff
@@ -29,6 +29,19 @@
         "System.Threading": "4.0.11-*",
         "System.Threading.Tasks": "4.0.11-*"
       }
+    },
+    "netcore50": {
+      "dependencies": {
+        "Microsoft.NETCore.Platforms": {
```
davidfowl: I'd like to get rid of this before we mrege this.

natemcmaster: We can't. All of these build dependencies are here to make dotnet-restore happy. With this, restore fails:

```
Errors in C:\Users\namc\dev\Universe\DependencyInjection\src\Microsoft.Extensions.DependencyInjection\project.json
    System.IO 4.0.0 provides a compile-time reference assembly for System.IO on .NETCore,Version=v5.0, but there is no
un-time assembly compatible with win7-x64.
    System.Runtime.InteropServices 4.0.0 provides a compile-time reference assembly for System.Runtime.InteropServices
n .NETCore,Version=v5.0, but there is no run-time assembly compatible with win7-x64.
    System.Text.Encoding 4.0.0 provides a compile-time reference assembly for System.Text.Encoding on .NETCore,Version
5.0, but there is no run-time assembly compatible with win7-x64.
    One or more packages are incompatible with .NETCore,Version=v5.0 (win7-x64).
    System.IO 4.0.0 provides a compile-time reference assembly for System.IO on .NETCore,Version=v5.0, but there is no
un-time assembly compatible with win7-x86.
    System.Runtime.InteropServices 4.0.0 provides a compile-time reference assembly for System.Runtime.InteropServices
n .NETCore,Version=v5.0, but there is no run-time assembly compatible with win7-x86.
    System.Text.Encoding 4.0.0 provides a compile-time reference assembly for System.Text.Encoding on .NETCore,Version
5.0, but there is no run-time assembly compatible with win7-x86.
    One or more packages are incompatible with .NETCore,Version=v5.0 (win7-x86).
```

PR #410, opened 2016-05-26
Use a poco type for verify ClosedServicesPreferredOverOpenGenericServices

### LINKED ISSUE #406: Property injection should fail for incomplete registrations
In the process of getting [LightInject ](www.lightinject.net) ready for RC2, I ran the tests against the latest bits and discovered that I had one failing test (ClosedServicesPreferredOverOpenGenericServices). This test used to pass and the only thing that has changed was it was using **AnotherClass** as the generic argument instead of a **string** in the old version. I soon discovered that it failed because that **AnotherClass** has a constructor dependency (IFakeService) that is not registered with the container. I understand the loose strategy for property injection, but IMHO it should fail if the property dependency is registered and that dependency in turn has a dependency that is not registered/resolvable. Otherwise this could lead to some scenarios that could be very hard to debug. 

Let me illustrate what I mean with a couple of tests.

```
 public void ShouldHandleMissingPropertyDepenency()
{
    // Arrange 
    var collection = new ServiceCollection();
    collection.AddTransient(typeof(FakeService), typeof(FakeService));
    var provider = CreateServiceProvider(collection);
    
    // Act 
    var service = provider.GetService<FakeService>();
    
    // Assert
    Assert.IsType<FakeService>(service);
    Assert.Null(service.Value);                        
}
```

The fact that the value is null(AnotherClass) is perfectly fine since it was never registered into the service collection. The next test demonstrates the problem.

```
public void ShouldThrowExceptionForIncompletePropertyDependency()
{           
    // Arrange
    var collection = new ServiceCollection();
    collection.AddTransient(typeof(FakeService));
    collection.AddTransient(typeof(AnotherClass));
    var provider = CreateServiceProvider(collection);
    
    // THIS SHOULD FAIL 
    var service = provider.GetService<FakeService>();
          
}
```

This test should fail because the property dependency (AnotherClass) is registered, but is not actually resolvable because AnotherClass has a unregistered required constructor dependency.

If this (dangerous) behavior is by design, there should at least be tests that verifies this behavior.

pranavkm:
Fixes #406

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/DependencyInjectionSpecificationTests.cs
+++ b/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/DependencyInjectionSpecificationTests.cs
@@ -536,13 +536,13 @@ namespace Microsoft.Extensions.DependencyInjection.Specification
         {
             // Arrange
             var collection = new ServiceCollection();
-            collection.AddTransient(typeof(IFakeOpenGenericService<AnotherClass>), typeof(FakeService));
+            collection.AddTransient(typeof(IFakeOpenGenericService<PocoClass>), typeof(FakeService));
             collection.AddTransient(typeof(IFakeOpenGenericService<>), typeof(FakeOpenGenericService<>));
-            collection.AddSingleton<AnotherClass>();
+            collection.AddSingleton<PocoClass>();
             var provider = CreateServiceProvider(collection);
 
             // Act
-            var service = provider.GetService<IFakeOpenGenericService<AnotherClass>>();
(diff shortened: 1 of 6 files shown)


pranavkm:
cc @Eilon \ @seesharper

javiercn:
:shipit:

Eilon:
:shipit:

seesharper:
Good stuff :shipit:

davidfowl:
:shipit:

seesharper:
Although this fixes the ClosedServicesPreferredOverOpenGenericServices test, it does not however fix the underlying problem with incomplete graphs being injected into properties. In my opinion there is no reason to close #406 just yet.

Issue #425, opened 2016-06-28
Avoid lambda expression compilation for Singleton scoped services


mnltejaswini:
The motivation for this issue is customer issue https://github.com/aspnet/Hosting/issues/781 
Customer reported that for his app on Desktop CLR (net 46) , there is an increase in the memory foot print from RC1 to RC2. But as of now we care about RTM. So I collected some numbers to compare memory between RC2 to RTM for the same app that customer shared with us. Here are the numbers after the first request and doing a full GC sweep.

| Metric   |      RC2      |  RTM |
|----------|:-------------:|------:|
| Total Private Bytes |  250 MB | 315 MB |
| Total Managed Heap Private Bytes |    123 MB   |   180 MB |

From the numbers,  there is an increase in the memory from RC2 to RTM during start up of the app. The increase comes from managed heap only.  Although there isn't increase in the retained memory in GC , there is an increase of 100MB of memory allocations during start up of the app. The dotMemory profile for the allocations is pasted below

RC2 memory allocations during start up

![image](https://cloud.githubusercontent.com/assets/8011073/16424696/18ad3a06-3d16-11e6-92f2-79ac39e33b4e.png)

RTM memory allocations during start up 
![image](https://cloud.githubusercontent.com/assets/8011073/16424750/50d0a56c-3d16-11e6-95f8-81bb98841dd2.png)

As seen, the major memory allocations come from Code Access Security checks that are executed during compilation of lambda expression when a service is resolved more than 2 times. (https://github.com/aspnet/DependencyInjection/blob/dev/src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs#L86 )  With the increase in the number of services resolved, these allocations increase.
However Net core apps doesn't have these checks as they are removed for Core CLR.
Note that these CAS objects also appear in the top list for RC2. Although these allocations happens one time, the problem is that the GC managed heap shoots high triggering GC cycles. Also, this increases the memory foot print for the process. This issue is logged to reduce these memory allocations during start up.  
Singleton services once created are always available in the lookup dictionary. So it is cheap to  directly call the Invoke method that does a simple lookup rather than building expression trees to do the same. I prototyped this change and measured the memory.  In the customer app, there are about 903 Singleton services, 38 scoped services and 463 transient services. As there are more Singleton services, the change has very good impact. Below are the numbers

| Metric   |      Before       |  After |
|----------|:-------------:|------:|
| Total Private Bytes |  315MB | 223MB |
[...]

PR #426, opened 2016-06-28
Perf: Avoid lambda expression compilation for Singleton scoped services

mnltejaswini:
The complete details for this PR are included in the #425.

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
@@ -73,7 +73,12 @@ namespace Microsoft.Extensions.DependencyInjection
             var callCount = 0;
             return provider =>
             {
-                if (Interlocked.Increment(ref callCount) == 2)
+                // PERF: Singleton services once created will always be available in the lookup dictionary.
+                // Hence it is not needed to create an lambda expression for looking up in the dictionary.
+                // This change reduces the memory allocations during the start up because 
+                // lambda expression compilation involves expensive code access security checks for Desktop CLR.
+                var singletonCallSite = callSite as SingletonCallSite;
+                if (singletonCallSite ==  null && Interlocked.Increment(ref callCount) == 2)
                 {
                     Task.Run(() =>
                     {
@@ -214,6 +219,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
         private static MethodInfo MonitorEnterMethodInfo = GetMethodInfo<Action<object, bool>>((lockObj, lockTaken) => Monitor.Enter(lockObj, ref lockTaken));
         private static MethodInfo MonitorExitMethodInfo = GetMethodInfo<Action<object>>(lockObj => Monitor.Exit(lockObj));
+        private static MethodInfo InvokeMethodInfo = GetMethodInfo<Func<IServiceCallSite, ServiceProvider, object>>((a, b) => a.Invoke(b));
 
         private static MethodInfo GetMethodInfo<T>(Expression<T> expr)
         {
@@ -362,7 +368,15 @@ namespace Microsoft.Extensions.DependencyInjection
 
             public override Expression Build(Expression provider)
             {
-                return base.Build(Expression.Field(provider, "_root"));
+                // PERF: Singleton services once created will always be available in the lookup dictionary
+                // Hence it is cheaper to directly call its Invoke method that will do the lookup rather than 
+                // building an expression tree that does the same.
+                // This change reduces the memory allocations during the start up because, expressions accessing private members
+                // involves expensive code access security checks for Desktop CLR.
+                return Expression.Call(
+                    Expression.Constant(this, typeof(SingletonCallSite)),
+                    InvokeMethodInfo,
+                    provider);
             }
         }
     }


mnltejaswini:
@rynowak

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
```diff
@@ -73,7 +73,12 @@ public object GetService(Type serviceType)
             var callCount = 0;
             return provider =>
             {
-                if (Interlocked.Increment(ref callCount) == 2)
+                // PERF: Singleton services once created will always be available in the lookup dictionary.
+                // Hence it is not needed to create an lambda expression for looking up in the dictionary.
+                // This change reduces the memory allocations during the start up because 
+                // lambda expression compilation involves expensive code access security checks for Desktop CLR.
+                var singletonCallSite = callSite as SingletonCallSite;
```
pranavkm: Would there be any benefit in not creating the `Func` for singletons? i.e. have `RealizedServices` store the instance when it's a singleton and `Func` otherwise?

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
```diff
@@ -362,7 +368,15 @@ public override object Invoke(ServiceProvider provider)
...
-                return base.Build(Expression.Field(provider, "_root"));
+                // PERF: Singleton services once created will always be available in the lookup dictionary
+                // Hence it is cheaper to directly call its Invoke method that will do the lookup rather than 
+                // building an expression tree that does the same.
+                // This change reduces the memory allocations during the start up because, expressions accessing private members
+                // involves expensive code access security checks for Desktop CLR.
+                return Expression.Call(
+                    Expression.Constant(this, typeof(SingletonCallSite)),
+                    InvokeMethodInfo,
+                    provider);
```
rynowak: This still appears to involve generating an expression - why is that needed?

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
rynowak: That would double the size of the service table.

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
mnltejaswini: Yes. This change is for other call site types that reference singletons

pranavkm: Also make a note that the singleton would have already been registered for disposal at this point. (We're removing the call to `base.Build`).

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
pranavkm: Didn't see a particularly cleaner pattern removing the polymorphism. It's used as a way to reuse code here and changing it would just involve a lot of code re-organization without any specific benefits. That said, it would make sense to move this check to `CreateServiceAccessor` and special case top level singletons outside the closure.

lodejard: The only changes would be for `IServiceProvider.GetService` directly for a singleton. Plus it happens out-of-band (in a Task, not in the main-line code path). Is this a realistic optimization?

rynowak: Expression compilation allocates to an extreme degree on desktop due to CAS. https://github.com/aspnet/DependencyInjection/issues/425 This is causing a lot of heap-size growth at startup. As far as the breakdown between the impact of this change vs the change to `SingletonCallSite.Build`, @mnltejaswini will have to answer

lodejard: Okay... Then how about moving the down-cast into the Task.Run body? if callSite is singleton, `table.RealizedService[serviceType] = callSite.Invoke;` else compile expression?

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
lodejard: Is the memory allocation proportional to the number of expression nodes, or to the number of calls to lambda compile?

Assuming it's proportional to the number of calls to compile - I don't think you need to change this method at all if the top-level Build isn't called for direct calls to GetService of singletons.

mnltejaswini: Although avoiding top level Build optimizes the direct calls to GetService of singletons, this change to Build optimizes the GetService of other call sites that have singletons in their expression graph. I will get some numbers for each of the change separately for the customer app.

mnltejaswini: Here are the numbers measured for startup with each change separately. The maximum benefit comes from changing the Build method of SIngletonCallSite

| Metric   |      Total allocations       |  Total private bytes of managed Heap |
|----------|:-------------:|------:|
| Avoid Build for direct calls to get Singleton services |  239MB | 145MB |
| Changing Build for Singleton call site to reduce the nodes in the expression tree for dependent services|    115MB   |   95 MB |

davidfowl: To @pranavkm's point about singletons being added to the list of disposables, how does this work in those cases? Is it because invoke gets called before this ever does?

halter73: This actually looks pretty safe and reasonable. Calling `Invoke()` instead of doing the equivalent action directly in the expression tree isn't really that costly since we'll always be pulling the instance out of a dictionary anyway instead of doing any complicated reflection, constructor selection, etc... If calling `Invoke()` avoids costly CAS-related allocations, that seems better to me than some other sugested alternatives such as making `_root` public.

@davidfowl The new compiled expression returned by this method, `SingletonCallSite.Build`, calls into `ScopedCallSite.Invoke` which would add a newly created singleton to `_resolvedServices` (used in part to track disposables) if it wasn't there already. Even though it's not really possible as @pranavkm mentions for the singleton not to be in the dictionary yet, the code would still be correct (albeit slow) if that wasn't the case.

halter73:
:shipit: This doesn't preclude doing me extensive changes (e.g. memoization) later, and it certainly seems way better than what we have now on desktop.

pakrym:
This PR fixes allocations for singletons but not scoped services, avoiding CAS checks when generating expressions would fix both.

rynowak:
Per discussion, @pakrym is going to follow up on this, and possibly produce a more all-encompassing fix.

pakrym:
Closing. Issue this PR addressed was fixed in https://github.com/aspnet/DependencyInjection/pull/429

PR #429, opened 2016-07-12
Optimize expression compilation memory usage

pakrym:
Fixes: https://github.com/aspnet/DependencyInjection/issues/425 by caching private field access into local variables.
Also optimizes amount of `Monitor.Enter/Exit` and `try/catch` we do for scope services.
## Test case:

IdentitySample.Mvc project
Run `net451` and request `/` multiple times to force all services to compile.
## Before

**234 MB** allocation from `Microsoft.Extensions.DependencyInjection.ServiceProvider+<>c__DisplayClass17_0.<RealizeService>b__1()`, **223 MB** because of CAS checks.
And **2700ms** total compilation time.

## After

**9.8 MB** allocation from `Microsoft.Extensions.DependencyInjection.ServiceProvider+<>c__DisplayClass17_0.<RealizeService>b__1()`, **4.3 MB** because of CAS checks.
And **400ms** total compilation time.

## After + No trees for singletons

**2.7 MB** allocation from `Microsoft.Extensions.DependencyInjection.ServiceProvider+<>c__DisplayClass17_0.<RealizeService>b__1()`, **1.8 MB** because of CAS checks.
And **122ms** total compilation time.

## After + No trees for singletons + Short circuit singletons

https://github.com/aspnet/DependencyInjection/pull/429/files#diff-800d14c6aa47e810955d46f860d9502cR52

**1.6 MB** allocation from `Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteExpressionBuilder.Build(IServiceCallSite)`, **1.2 MB** because of CAS checks.
And **89ms** total compilation time.

@halter73 @davidfowl

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
@@ -17,6 +17,8 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
         private static readonly MethodInfo AddMethodInfo = GetMethodInfo<Action<IDictionary<object, object>, object, object>>((a, b, c) => a.Add(b, c));
         private static readonly MethodInfo MonitorEnterMethodInfo = GetMethodInfo<Action<object, bool>>((lockObj, lockTaken) => Monitor.Enter(lockObj, ref lockTaken));
         private static readonly MethodInfo MonitorExitMethodInfo = GetMethodInfo<Action<object>>(lockObj => Monitor.Exit(lockObj));
+        private static readonly MethodInfo CallSiteRuntimeResolverResolve =
+            GetMethodInfo<Func<CallSiteRuntimeResolver, IServiceCallSite, ServiceProvider, object>>((r, c, p) => r.Resolve(c, p));
 
         private readonly ParameterExpression _providerParameter = Expression.Parameter(typeof(ServiceProvider));
 
@@ -27,7 +29,16 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
         private readonly Dictionary<Expression, ParameterExpression> _resolvedServices = new Dictionary<Expression, ParameterExpression>();
         private readonly Dictionary<Expression, LambdaExpression> _captureDisposable = new Dictionary<Expression, LambdaExpression>();
 
-        private ParameterExpression _rootVariable;
+        private readonly CallSiteRuntimeResolver _runtimeResolver;
+
+        public CallSiteExpressionBuilder(CallSiteRuntimeResolver runtimeResolver)
+        {
+            if (runtimeResolver == null)
+            {
+                throw new ArgumentNullException(nameof(runtimeResolver));
+            }
+            _runtimeResolver = runtimeResolver;
+        }
 
         public Expression<Func<ServiceProvider, object>> Build(IServiceCallSite callSite)
         {
@@ -48,7 +59,15 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
 
         protected override Expression VisitSingleton(SingletonCallSite singletonCallSite, ParameterExpression provider)
         {
-            return VisitScoped(singletonCallSite, GetRootProvider());
+            // Call to CallSiteRuntimeResolver.Resolve is being returned here
+            // because in the current use case singleton service was already resolved and cached
+            // to dictionary so there is no need to generate full tree at this point.
+
+            return Expression.Call(
+                Expression.Constant(_runtimeResolver),
+                CallSiteRuntimeResolverResolve,
+                Expression.Constant(singletonCallSite, typeof(IServiceCallSite)),
+                provider);
         }
 
         protected override Expression VisitConstant(ConstantCallSite constantCallSite, ParameterExpression provider)
@@ -197,19 +216,6 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
             }
 
             return resolvedServicesVariable;
-        }
-
-        public ParameterExpression GetRootProvider()
-        {
-            if (_rootVariable == null)
-            {
-                var rootExpression = Expression.Property(_providerParameter, nameof(ServiceProvider.Root));
-                _rootVariable = Expression.Variable(typeof(ServiceProvider), "root");
-
-                _variables.Add(_rootVariable);
-                _initializations.Add(Expression.Assign(_rootVariable, rootExpression));
-            }
-            return _rootVariable;
         }
 
         private static Expression Lock(Expression body, Expression syncVariable)
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
@@ -1,0 +1,260 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Threading;
+
+namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
+{
+    internal class CallSiteExpressionBuilder
+    {
+        private static readonly MethodInfo CaptureDisposableMethodInfo = GetMethodInfo<Func<ServiceProvider, object, object>>((a, b) => a.CaptureDisposable(b));
+        private static readonly MethodInfo TryGetValueMethodInfo = GetMethodInfo<Func<IDictionary<object, object>, object, object, bool>>((a, b, c) => a.TryGetValue(b, out c));
+        private static readonly MethodInfo AddMethodInfo = GetMethodInfo<Action<IDictionary<object, object>, object, object>>((a, b, c) => a.Add(b, c));
+        private static readonly MethodInfo MonitorEnterMethodInfo = GetMethodInfo<Action<object, bool>>((lockObj, lockTaken) => Monitor.Enter(lockObj, ref lockTaken));
+        private static readonly MethodInfo MonitorExitMethodInfo = GetMethodInfo<Action<object>>(lockObj => Monitor.Exit(lockObj));
+
+        private readonly ParameterExpression _providerParameter = Expression.Parameter(typeof(ServiceProvider));
+
+        private readonly List<Expression> _initializations = new List<Expression>();
+        private readonly List<ParameterExpression> _variables = new List<ParameterExpression>();
+        private readonly List<Expression> _locks = new List<Expression>();
+
+        private readonly Dictionary<Expression, ParameterExpression> _resolvedServices = new Dictionary<Expression, ParameterExpression>();
+        private readonly Dictionary<Expression, LambdaExpression> _captureDisposable = new Dictionary<Expression, LambdaExpression>();
+
+        private ParameterExpression _rootVariable;
+
+        public Expression<Func<ServiceProvider, object>> Build(IServiceCallSite callSite)
+        {
+            var serviceExpression = BuildExpression(callSite, _providerParameter);
+
+            List<Expression> body = new List<Expression>();
+            body.AddRange(_initializations);
+            foreach (var expression in _locks)
+            {
+                serviceExpression = Lock(serviceExpression, expression);
+            }
+            body.Add(serviceExpression);
+
+            return Expression.Lambda<Func<ServiceProvider, object>>(
+                Expression.Block(_variables, body),
+                _providerParameter);
+        }
+
+        private Expression BuildExpression(IServiceCallSite callSite, ParameterExpression provider)
+        {
+            var factoryService = callSite as FactoryService;
+            if (factoryService != null)
+            {
+                return BuildFactoryService(factoryService, provider);
+            }
+            var closedIEnumerableCallSite = callSite as ClosedIEnumerableCallSite;
+            if (closedIEnumerableCallSite != null)
+            {
+                return BuildClosedIEnumerable(closedIEnumerableCallSite, provider);
+            }
+            var constructorCallSite = callSite as ConstructorCallSite;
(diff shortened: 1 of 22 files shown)


src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
```diff
@@ -17,18 +17,19 @@ namespace Microsoft.Extensions.DependencyInjection
...
     internal class ServiceProvider : IServiceProvider, IDisposable
     {
-        private readonly ServiceProvider _root;
         private readonly ServiceTable _table;
         private bool _disposeCalled;
-
-        private readonly Dictionary<IService, object> _resolvedServices = new Dictionary<IService, object>();
         private List<IDisposable> _transientDisposables;
 
+        internal ServiceProvider Root { get; }
```
halter73: Do we generally use `internal` or `public` on classes that are already `internal`?

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/IServiceCallSite.cs
```diff
@@ -1,17 +1,12 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Linq.Expressions;
-
 namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
 {
     /// <summary>
     /// Summary description for IServiceCallSite
     /// </summary>
     internal interface IServiceCallSite
```
halter73: Marker interface? Am I correct in assuming you could change all instances of `IServiceCallSite` to `object` and the code would still compile and work?

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        private ParameterExpression _rootVariable;
+
+        public Expression<Func<ServiceProvider, object>> Build(IServiceCallSite callSite)
+        {
+            var serviceExpression = VisitCallSite(callSite, _providerParameter);
+
+            List<Expression> body = new List<Expression>();
+            body.AddRange(_initializations);
+            foreach (var expression in _locks)
```
halter73: In the current code, doesn't `_locks` contain either zero or one element, the `resolvedServicesVariable`. That's the only thing `ServiceProvider` ever locks on.

BTW, since we are changing the code to only lock once for access to `_resolvedServices`, you might as well do the same for `_transientDisposables`. This would mean not calling the current `CaptureDisposable` in the generated expression which takes the lock.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
```diff
@@ -0,0 +1,233 @@
...
+            body.Add(serviceExpression);
+
+            return Expression.Lambda<Func<ServiceProvider, object>>(
+                Expression.Block(_variables, body),
+                _providerParameter);
+        }
+
+        protected override Expression VisitSingleton(SingletonCallSite singletonCallSite, ParameterExpression provider)
+        {
+            return VisitScoped(singletonCallSite, GetRootProvider());
```
halter73: I'm no longer sold on building complicated expression trees for singletons, since it always winds up being one dictionary lookup once the tree is built in practice anyways. Wouldn't it be better to avoid CAS altogether for the root `_resolvedServices` by creating an expression that calls `CallSiteRuntimeResolver.Resolve(singletonCallSite, provider);` like #426.

halter73:
@SajayAntony Do you still have the application mentioned in https://github.com/aspnet/Hosting/issues/781 that was used in testing the impact of #426?

halter73:
I really like the design of this btw @pakrym :+1:

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/IServiceCallSite.cs
pakrym: Yeah, just to make sure we don't start putting random objects by accident.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
pakrym: I think it can be either `provider._resolvedServices` or `provider._root._resolvedServices` and if there is a scoped service and a singleton in same expression we can get two locks.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
pakrym: I see your point, my though was that relying on other class behavior (that expression would be compiled after being resolved at least once)  is not very clean approach.

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
pakrym: I don't have strong opinion on this, `internal` property in `internal` class seems to be pointless now when I look at it.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
pakrym: `CaptureDisposable` has more logic then just lock which I don't really want to move to `Expressions` (lazy init, reference equality checks)

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
```diff
@@ -0,0 +1,233 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Threading;
+
+namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
+{
+    internal class CallSiteExpressionBuilder: CallSiteVisitor<ParameterExpression, Expression>
```
davidfowl: nit: formatting

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteRuntimeResolver.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        protected override object VisitSingleton(SingletonCallSite singletonCallSite, ServiceProvider provider)
+        {
+            return VisitScoped(singletonCallSite, provider.Root);
+        }
+
+        protected override object VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
+        {
+            object resolved;
+            lock (provider.ResolvedServices)
```
davidfowl: should this be `lock (provider.SyncRoot)`

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteRuntimeResolver.cs
```diff
@@ -0,0 +1,116 @@
+using System;
+using System.Runtime.ExceptionServices;
+
+namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
+{
+    class CallSiteRuntimeResolver : CallSiteVisitor<ServiceProvider, object>
```
davidfowl: nit: internal

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
```diff
@@ -17,18 +17,19 @@ namespace Microsoft.Extensions.DependencyInjection
...
+        internal Dictionary<object, object> ResolvedServices { get; } = new Dictionary<object, object>();
+
         private static readonly Func<Type, ServiceProvider, Func<ServiceProvider, object>> _createServiceAccessor = CreateServiceAccessor;
+        private static CallSiteRuntimeResolver _callSiteRuntimeResolver = new CallSiteRuntimeResolver();
```
davidfowl: Add a comment saying that this is completely stateless so it's safe to cache (unlike the expression generator)

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
```diff
@@ -164,16 +162,16 @@ public void Dispose()
                 // PERF: We've enumerating the dictionary so that we don't allocate to enumerate.
                 // .Values allocates a KeyCollection on the heap, enumerating the dictionary allocates
```
rynowak: This isn't really true `.Values` allocates a `ValueCollection` :trollface:

davidfowl: I think that's my comment 😄

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
davidfowl: It's not but it's practical with a giant comment about why the code is that way and what the savings are (if there are much more past what you got already)

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
davidfowl: A comment here would be useful.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteRuntimeResolver.cs
pakrym: Previous generated code locked on `ResolvedServices`, so I kept it that way too.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
pakrym: We can go even further and generate constant expression with result of `CallSiteRuntimeResolver.Resolve`

pakrym: No we can't, there is no service provider at that point.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
halter73: Since `CaptureDisposable` doesn't ever call back into a compiled expression, I don't think it's too important to translate it into an expression tree. I was more suggesting a `LocklessCaptureDisposable` that could be called if a lock is already taken at the top level (which it likely would be for scoped/singleton services. Of course, it only really matters if re-entering the lock is costly, which even at this level it probably not. That's why I phrased it as "if we're doing this anyway for other things..."

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteRuntimeResolver.cs
halter73: Let's consolidate this. This also threw me off when reviewing and searching the class for `lock (` and seeing two expressions that are really the same. Use `SyncRoot` everywhere or remove it.

pakrym: This would result in two CAS checked property access instead one: one for `SyncRoot` other for `ResolvedServices`, we can remove `SyncRoot` and lock onto `ResolvedServices and `_transientDisposables` individually in service provider.

davidfowl: Lets remove it then

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteVisitor.cs
```diff
@@ -0,0 +1,96 @@
...
+            if (emptyIEnumerableCallSite != null)
+            {
+                return VisitEmptyIEnumerable(emptyIEnumerableCallSite, argument);
+            }
+            var serviceScopeService = callSite as ServiceScopeService;
+            if (serviceScopeService != null)
+            {
+                return VisitServiceScopeService(serviceScopeService, argument);
+            }
+            throw new NotSupportedException("Call site type is not supported");
```
halter73: In the absence of proper pattern matching, let's change this to `$"Call sit type '{callSite.GetType()}' is not supported"` so we don't have to attach a debugger when we mess up.

halter73:
Now I want to know if this makes things faster on coreclr. The smaller trees and avoiding compilation for singletons could be helpful even there.

:shipit:

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
```diff
@@ -0,0 +1,242 @@
...
+                return (provider) => _runtimeResolver.Resolve(callSite, provider);
+            }
+            return BuildExpression(callSite).Compile();
+        }
+
+        private Expression<Func<ServiceProvider, object>> BuildExpression(IServiceCallSite callSite)
+        {
+            var serviceExpression = VisitCallSite(callSite, ProviderParameter);
+
+            List<Expression> body = new List<Expression>();
```
davidfowl: var

SajayAntony:
@halter73. I've shared the app with you and @pakrym

pakrym:
Customers app

## Before:


## After

MaherJendoubi:
@pakrym Which tool are you using?

halter73:
@MaherJendoubi dotMemory

Issue #88, opened 2014-06-06
Throw when accessing a scoped service from a non-scoped service


rynowak:
We hit an instance of this in MVC - see aspnet/Mvc#631

`UrlHelper` is a scoped service - and an attempt was made to resolve it from inside a singleton.

Can we throw if this happens?

PR #430, opened 2016-07-14
Validate scope into singleton sevice injection

pakrym:
Fixes:https://github.com/aspnet/DependencyInjection/issues/88
@davidfowl @halter73


src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
```diff
@@ -9,7 +9,12 @@ public static class ServiceCollectionContainerBuilderExtensions
     {
         public static IServiceProvider BuildServiceProvider(this IServiceCollection services)
         {
-            return new ServiceProvider(services);
+            return BuildServiceProvider(services, false);
```
davidfowl: nit: style `validateScopes: false`

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/IService.cs
```diff
@@ -13,5 +13,7 @@ internal interface IService
         ServiceLifetime Lifetime { get; }
 
         IServiceCallSite CreateCallSite(ServiceProvider provider, ISet<Type> callSiteChain);
+
+        Type Type { get; }
```
davidfowl: ServiceType?

davidfowl:
Make that style change where the optional parameter syntax is used everywhere

davidfowl:
I think there's value in being able to catch this at startup time before the first resolve.

pakrym:
Indeed there is but service provider is lazy for a reason and some of them could not be caught like open generics use cases `ISingleton<T>` where `T is IHaveScopedService`

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteValidatorState.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
+{
+    internal struct CallSiteValidatorState
```
pranavkm: Make this nested inside `CallSiteValidator`? It's not used outside of it.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteValidatorState.cs
```diff
@@ -0,0 +1,10 @@
...
+    {
+        public SingletonCallSite Singleton;
```
pranavkm: Property? We generally don't use fields.

davidfowl: It's an internal struct, maybe it should be a private instead

test/Microsoft.Extensions.DependencyInjection.Tests/CallSiteTests.cs
```diff
@@ -82,7 +82,7 @@ public static IEnumerable<object[]> TestServiceDescriptors(ServiceLifetime lifet
         public void BuiltExpressionWillReturnResolvedServiceWhenAppropriate(
             ServiceDescriptor[] desciptors, Type serviceType, Func<object, object, bool> compare)
         {
-            var provider = new ServiceProvider(desciptors);
+            var provider = new ServiceProvider(desciptors, validateScopes: false);
```
pranavkm: Why do these tests not like the validation? Could we just fix them instead

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderValidationTests.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
```
pranavkm: License

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderValidationTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+        }
+
+        private class Boo : IBoo
+        {
+            public Boo(IServiceScopeFactory scopeFactory)
+            {
+            }
+        }
+
```
pranavkm: Empty line.

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderValidationTests.cs
```diff
@@ -0,0 +1,65 @@
...
+namespace Microsoft.Extensions.DependencyInjection.Tests
+{
+    public class ServiceProviderValidationTests
+    {
+        [Fact]
+        public void GetService_Throws_WhenScopedIsInjectedIntoSingleton()
+        {
+            // Arrange
+            var serviceCollection = new ServiceCollection();
```
pranavkm: Could we get tests with a couple more variants

* IFoo (Singleton) -> IBar (Singleton) ->  IBaz (Scoped)
* IFoo (Singleton) -> IBar (Transient) ->  IBaz (Scoped)

src/Microsoft.Extensions.DependencyInjection/Resources.resx
```diff
@@ -141,4 +141,7 @@
     <value>A suitable constructor for type '{0}' could not be located. Ensure the type is concrete and services are registered for all parameters of a public constructor.</value>
     <comment>{0} = service type</comment>
   </data>
+  <data name="ScopedInSingletonException" xml:space="preserve">
+    <value>Cannot consume scoped service '{0}' from singleton '{1}'.</value>
```
pranavkm: "Service '{0}' with {1} lifetime cannot consume service '{2}' with {3} lifetime."? and use `nameof(ServiceLifetime.Singleton)`

pranavkm:
:watch: for the few questions.

src/Microsoft.Extensions.DependencyInjection/Resources.resx
pakrym: It's always singleton, I'm not sure if this add any value.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteValidatorState.cs
pranavkm: Wouldn't work since it's part of the interface contract.

src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
```diff
@@ -9,7 +9,12 @@ public static class ServiceCollectionContainerBuilderExtensions
     {
         public static IServiceProvider BuildServiceProvider(this IServiceCollection services)
         {
-            return new ServiceProvider(services);
+            return BuildServiceProvider(services, validateScopes: false);
+        }
+
+        public static IServiceProvider BuildServiceProvider(this IServiceCollection services, bool validateScopes)
```
halter73: Time for doc comments? On one hand, I kind of hope no one ever uses this (except for may us). On the other hand, without context, I would have no idea what `validateScopes` is even about.

halter73:
Is the issue we want to identify really a scoped service being injected into a singleton (perhaps transitively). Or is the issue even more fundamental, should we vialidate that a scoped service is never resolved from the root container? We might have to validate on every resolve for that, but it should be simple.

src/Microsoft.Extensions.DependencyInjection/Resources.resx
pranavkm: Well, in general you want to avoid putting code "things" as part of loc comments (since there's no way to localize them). In this case, the lifetime names are enum values and don't benefit from being localized.

src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
pranavkm: For the other method + type too since they're all public.

pakrym:
:up::date:

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteValidator.cs
```diff
@@ -0,0 +1,114 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
+{
+    internal class CallSiteValidator: CallSiteVisitor<CallSiteValidator.CallSiteValidatorState, Type>
+    {
+        private readonly Dictionary<Type, Type> _scopedServices = new Dictionary<Type, Type>();
```
halter73: Comment to indicate the key is the service being resolved via `GetService` and the value is a single scoped service that it depends on (but there may be more.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteValidator.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        public void ValidateResolution(Type serviceType, ServiceProvider serviceProvider)
+        {
+            Type scopedService;
+            if (ReferenceEquals(serviceProvider, serviceProvider.Root)
+                && _scopedServices.TryGetValue(serviceType, out scopedService))
+            {
+                throw new InvalidOperationException(Resources.FormatScopedResolvedFromRootException(
+                    serviceType,
+                    scopedService,
```
halter73: Can't `serviceType` and `scopedService` be the same? That would make the message unnecessarily confusing.

src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
```diff
@@ -7,9 +7,28 @@ namespace Microsoft.Extensions.DependencyInjection
 {
     public static class ServiceCollectionContainerBuilderExtensions
     {
+        /// <summary>
+        /// Creates an <see cref="IServiceProvider"/> containing services from the provided <see cref="IServiceCollection"/>.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/> containing service descriptors.</param>
+        /// <returns>An instance of <see cref="IServiceProvider"/> providing access to services.</returns>
```
pranavkm: `<returns>The <see cref="IServiceProvider"/>.</returns>`

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderValidationTests.cs
```diff
@@ -0,0 +1,128 @@
...
+        public void GetService_Throws_WhenGetServiceForScopedServiceIsCalledOnRoot()
+        {
+            // Arrange
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddTransient<IFoo, Foo>();
+            serviceCollection.AddScoped<IBar, Bar>();
+            var serviceProvider = serviceCollection.BuildServiceProvider(validateScopes: true);
+
+            // Act + Assert
+            var exception = Assert.Throws<InvalidOperationException>(() => serviceProvider.GetService(typeof(IFoo)));
```
halter73: Might as well assert `serviceProvider.GetService(typeof(IBar))` also throws.

halter73:
:shipit:

src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
```diff
@@ -7,9 +7,28 @@ namespace Microsoft.Extensions.DependencyInjection
...
+            return BuildServiceProvider(services, validateScopes: false);
+        }
+
+        /// <summary>
+        /// Creates an <see cref="IServiceProvider"/> containing services from the provided <see cref="IServiceCollection"/>
+        /// optionaly enabling scope validation.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/> containing service descriptors.</param>
+        /// <param name="validateScopes">
+        /// <c>true</c> to perform check verifying that scoped services never gets resolved from root provider; otherwise <c>false</c>.
```
halter73: Might as well point out `false` is the default behavior.

pranavkm:
:shipit:

khellang:
In relation to @halter73's comment; I've always wondered what happens if/when you resolve a scoped service from the top-level container... Does it effectively become a singleton?

khellang:
If that's the case, you have (at least) two ways a scoped/transient service can be held captive; either as a dependency of a singleton/scoped service or in the top-level provider.

pakrym:
@khellang yes, this PR is targeting these two scenarios.

halter73:
:ship:

cwe1ss:
This is a great addition! Since it is now defined that resolving scoped services from a non-scoped provider should not be possible, shouldn't this also be covered in the Specification.Tests library to make sure other containers don't allow this either?

Issue #442, opened 2016-08-05
Proposal: Add IServiceProviderFactory


davidfowl:
This will allow 3rd party containers to plug into the default pipeline when the application doesn't:

```C#
public interface IServiceProviderFactory<TContainerBuilder>
{
    TContainerBuilder CreateContainerBuilder(IServiceCollection services);
    IServiceProvider CreateServiceProvider(TContainerBuilder containerBuilder);
}

public class ExampleAutofacServiceProviderFactory : IServiceProviderFactory<ContainerBuilder>
{
    public override ContainerBuilder CreateContainerBuilder(IServiceCollection services)
    {
         var containerBuilder = new ContainerBuilder();
         containerBuilder.Populate(services);
         return containerBuilder;
     }

    public override IServiceProvider CreateServiceProvider(ContainerBuilder containerBuilder)
    {
        return new AutofacServiceProvider(containerBuilder.Build());
    }
}
```

See https://github.com/aspnet/Hosting/issues/829 for consumption

/cc @tillig @khellang

khellang:
I like it! Reminds me a bit of some of the things I did with the Nancy bootstrapper prototype. This would cut down on the amount of code someone would have to write when you just want to use a third party container, but don't care about its registration APIs. You could question the need for a third party container in this case, though... But it looks like a nice addition.

tillig:
I like it, particularly in combination with the ConfigureServices overload. Getting a ContainerBuilder (or whatever your container takes) right into the method would save a lot of boilerplate (stuff I actually have in a few projects already).

davidfowl:
> You could question the need for a third party container in this case, though... But it looks like a nice addition.

Yea, that's why paired with the `ConfigureServices(Container c)` approach, it might be pretty nice.  

It would be great to get some more feedback here to see if this is a viable approach. We plan to use this ourselves to test out our system with different DI containers.

davidfowl:
I wonder if this should just be a hosting specific interface. To hang extension methods off the `IWebHost` you'd need to reference Hosting.Abstractions anyways. It'll matter more if other systems decide to use this interface for wire up.

khellang:
> I wonder if this should just be a hosting specific interface. To hang extension methods off the IWebHost you'd need to reference Hosting.Abstractions anyways.

[...]

PR #443, opened 2016-08-08
Add support for IServiceProviderFactory

davidfowl:
- A new abstraction that allows the DI container author to describe how
  to translate an IServiceCollection to a specific container and then to an IServiceProvider.
- Hosting will use this abstraction to swap out the default container

https://github.com/aspnet/DependencyInjection/issues/442

/cc @pakrym @halter73


src/Microsoft.Extensions.DependencyInjection.Abstractions/IServiceProviderFactory.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// Provides and extension point for creating a container specific builder and an <see cref="IServiceProvider"/>.
```
pranavkm: Provides an~~d~~ extension

src/Microsoft.Extensions.DependencyInjection.Abstractions/ServiceProviderFactory.cs
```diff
@@ -0,0 +1,34 @@
...
+        public abstract TContainerBuilder CreateBuilder(IServiceCollection services);
+
+        /// <summary>
+        /// Creates an <see cref="IServiceProvider"/> from the container builder.
+        /// </summary>
+        /// <param name="containerBuilder">The container builder</param>
+        /// <returns>An <see cref="IServiceProvider"/></returns>
+        public abstract IServiceProvider CreateServiceProvider(TContainerBuilder containerBuilder);
+
+        object IServiceProviderFactory.CreateBuilder(IServiceCollection services)
```
pranavkm: `object IServiceProviderFactory.CreateBuilder(IServiceCollection services) => CreateBuilder(services)`

for :100: points.

pranavkm:
Nice. :shipit: when everybody's happy.

src/Microsoft.Extensions.DependencyInjection.Abstractions/ServiceProviderFactory.cs
davidfowl: Long line

src/Microsoft.Extensions.DependencyInjection.Abstractions/ServiceProviderFactory.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// Provides an abstract base for creating a container specific builder and an <see cref="IServiceProvider"/>.
+    /// </summary>
+    public abstract class ServiceProviderFactory<TContainerBuilder> : IServiceProviderFactory
+    {
+        /// <summary>
+        /// Creates a container builder from an <see cref="IServiceCollection"/>.
+        /// </summary>
+        /// <param name="services">The collection of services</param>
+        /// <returns>A container builder that can be used to create an <see cref="IServiceProvider"/>.</returns>
+        public abstract TContainerBuilder CreateBuilder(IServiceCollection services);
```
khellang: `TBuilder`? How are you supposed to get to this type from `IServiceProvider`? Right now you have to call this method and do `builder.GetType()` in order to get it, but that might not be the exact type you'd want. Say you have a `ServiceProviderFactory<IContainerBuilder>`, but the actual type is `ContainerBuilder`. How will you find the `ConfigureContainer` method? Check for the parameter type using `IsAssignableFrom`?

khellang:
:shipit:

src/Microsoft.Extensions.DependencyInjection.Abstractions/ServiceProviderFactory.cs
davidfowl: Yep, using the runtime type of the return value.

src/Microsoft.Extensions.DependencyInjection.Abstractions/ServiceProviderFactory.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// Provides an abstract base for creating a container specific builder and an <see cref="IServiceProvider"/>.
+    /// </summary>
+    public abstract class ServiceProviderFactory<TContainerBuilder> : IServiceProviderFactory
```
pakrym: Why is this required?

davidfowl: It's not required. It's a convenient base class for people implementing it.

src/Microsoft.Extensions.DependencyInjection/DefaultServiceProviderFactory.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public class DefaultServiceProviderFactory : ServiceProviderFactory<IServiceCollection>
```
halter73: So this will be registered by hosting by default? I'm assuming we won't have an extension method like `WebHostBuilder.UseAutofac` for this.

davidfowl: > So this will be registered by hosting by default?

Yep

> I'm assuming we won't have an extension method like WebHostBuilder.UseAutofac for this.

Nope.

pakrym:
:shipit:

davidfowl:
Merged

Issue #420, opened 2016-06-16
Update Abstractions project.json description


NTaylorMullen:
The project.json for the DI abstractions [references](https://github.com/aspnet/DependencyInjection/blob/caa3f4ec3f0b7e92a0a0210795303b28c09ecbba/src/Microsoft.Extensions.DependencyInjection.Abstractions/project.json#L3) an old `ServiceCollectionExtensions` type.

This was renamed to `ServiceCollectionServiceExtensions`.

Eilon:
The fix is to remove that line entirely from the project.json file, because it's not a type that's commonly referenced from source code.

ajaybhargavb:
213a055708ea4b0031beb6a84334cbb2497e2496

PR #454, opened 2016-10-06
Removed description in project.json from DI Abstractions

ajaybhargavb:
https://github.com/aspnet/DependencyInjection/issues/420
@NTaylorMullen @Eilon


pranavkm:
We'll need to keep the first line of the description (remove the thing after the new line)

NTaylorMullen:
:shipit:

ajaybhargavb:
@Eilon just wants that particular type to be removed.

Issue #459, opened 2016-10-18
ServiceProvider should not dispose singleton instances


pakrym:
Right now it will dispose any `IDisposable` instance passed to AddSingleton when IServiceProvider is disposed.

PR #462, opened 2016-10-20
Do not dispose singleton instances

pakrym:
Fixes: https://github.com/aspnet/DependencyInjection/issues/459
@davidfowl @halter73

### DIFF
--- a/test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderContainerTests.cs
+++ b/test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderContainerTests.cs
@@ -101,9 +101,34 @@ namespace Microsoft.Extensions.DependencyInjection.Tests
                 exception.Message);
         }
 
+        [Fact]
+        public void DoesNotDisposeSingletonInstances()
+        {
+            var disposable = new Disposable();
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddSingleton(disposable);
+
+            var provider = serviceCollection.BuildServiceProvider();
+            provider.GetService<Disposable>();
+
+            ((IDisposable)provider).Dispose();
+
+            Assert.False(disposable.Disposed);
+        }
+
         private abstract class AbstractFakeOpenGenericService<T> : IFakeOpenGenericService<T>
         {
             public abstract T Value { get; }
+        }
+
+        private class Disposable : IDisposable
+        {
+            public bool Disposed { get; set; }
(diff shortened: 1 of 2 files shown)


test/Microsoft.Extensions.DependencyInjection.Tests/ServiceProviderContainerTests.cs
```diff
@@ -101,9 +101,34 @@ public void CreatingServiceProviderWithUnresolvableTypesThrows(Type serviceType,
...
+        public void DoesNotDisposeSingletonInstances()
+        {
+            var disposable = new Disposable();
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddSingleton(disposable);
+
+            var provider = serviceCollection.BuildServiceProvider();
+            provider.GetService<Disposable>();
+
+            (provider as IDisposable)?.Dispose();
```
halter73: Nit: hard cast

halter73 (approved):

davidfowl:
We need to see which of our types are going to start leaking because of this. I think the logger factory in hosting will need to be disposed manually.

davidfowl (approved):

khellang:
> We need to see which of our types are going to start leaking because of this.

Is this really a leak? The application container won't be disposed until the app shuts down anyway, right?

What about all the apps out there that might also "depend" on this behavior? Shouldn't this be classified as a breaking change? :thinking:

davidfowl:
> Is this really a leak? The application container won't be disposed until the app shuts down anyway, right?

Sure in the "default" application. Wait until people start hosting ASP.NET in a shared environment like visual studio. We shouldn't assume disposing the host means the application is done.

> What about all the apps out there that might also "depend" on this behavior? Shouldn't this be classified as a breaking change? 🤔

Yes, it is a breaking change.

/cc @DamianEdwards @Eilon

khellang:
> We shouldn't assume disposing the host means the application is done.

Good point :+1:

pakrym:
@davidfowl what about registering logger provider a as factory returning instance if we know it should be disposed with container?

Eilon:
FYI I marked the bug https://github.com/aspnet/DependencyInjection/issues/459 as breaking-change.

khellang:
You're making breaking changes to 1.1? :confused:

pakrym:
@khellang we are fine with taking this one because fixed behavior is much more useful and in typical application does not break anyone.

Issue #467, opened 2016-12-13
Race in ServiceLookup.CallSiteValidator.ValidateCallSite()


mikeharder:
@halter73, @pakrym 

I am getting the following exception when running the EF benchmarks (https://github.com/aspnet/benchmarks/issues/162):

```
System.ArgumentException: An item with the same key has already been added. Key: Benchmarks.Data.EfDb
   at System.ThrowHelper.ThrowAddingDuplicateWithKeyArgumentException(Object key)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.ValidateCallSite(Type serviceType, IServiceCallSite callSite) in C:\Git\DependencyInjection\src\Microsoft.Extensions.DependencyInjection\ServiceLookup\CallSiteValidator.cs:line 19
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.CreateServiceAccessor(Type serviceType, ServiceProvider serviceProvider) in C:\Git\DependencyInjection\src\Microsoft.Extensions.DependencyInjection\ServiceProvider.cs:line 76
   at System.Collections.Concurrent.ConcurrentDictionaryExtensions.GetOrAdd[TKey,TValue,TArg](ConcurrentDictionary`2 dictionary, TKey key, Func`3 valueFactory, TArg arg) in C:\Git\DependencyInjection\src\Microsoft.Extensions.DependencyInjection\Internal\ConcurrentDictionaryExtensions.cs:line 11
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType) in C:\Git\DependencyInjection\src\Microsoft.Extensions.DependencyInjection\ServiceProvider.cs:line 63
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetService[T](IServiceProvider provider) in C:\Git\DependencyInjection\src\Microsoft.Extensions.DependencyInjection.Abstractions\ServiceProviderServiceExtensions.cs:line 28
   at Benchmarks.Middleware.MultipleUpdatesEfMiddleware.<Invoke>d__4.MoveNext() in C:\Git\Benchmarks\src\Benchmarks\Middleware\MultipleUpdatesEfMiddleware.cs:line 0
```

The issue appears to be a race in ServiceLookup.CallSiteValidator.ValidateCallSite():

```
[...]

PR #470, opened 2016-12-21
Avoid race when scopeValidation is enabled

pakrym:
Fixes: https://github.com/aspnet/DependencyInjection/issues/467
@halter73 @mikeharder

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteValidator.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteValidator.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 
 namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteValidator.cs
```diff
@@ -2,21 +2,22 @@
...
-        private readonly Dictionary<Type, Type> _scopedServices = new Dictionary<Type, Type>();
+        private readonly ConcurrentDictionary<Type, Type> _scopedServices = new ConcurrentDictionary<Type, Type>();
 
         public void ValidateCallSite(Type serviceType, IServiceCallSite callSite)
         {
             var scoped = VisitCallSite(callSite, default(CallSiteValidatorState));
             if (scoped != null)
             {
-                _scopedServices.Add(serviceType, scoped);
+                _scopedServices[serviceType] = scoped;
```
mikeharder: Is it correct to update the value if the key already exists?  Or would it be more correct to use `TryAdd()`, so the value is *not* updated if the key already exists?  Or does it not matter and either would be equally correct?

mikeharder (changes requested):

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteValidator.cs
pakrym: We are displaying only one scoped service of multiple possible so I don't think there is a difference which one exactly.

mikeharder (approved):

DamianEdwards:
I don't know if there's any runway left for 1.1.1. @Eilon?

halter73:
Both a pro and con of merging this is that scopeValidation is opt-in. Presumably not many people are calling `collection.BuildServiceProvider(validateScopes: true)`.

davidfowl:
The problem is that there's no workaround if you do want to use it. That kinda bug IMO classifies as patchable. 1.1.2 maybe?

DamianEdwards:
Yeah I think it should be considered. I think it'll have to wait for 1.1.2 though

Eilon:
Too late for the current patch, for sure. Can someone log a new bug for DI's next milestone so that we can consider? Please include some reasonable justification in terms of customer scenario.

Issue #476, opened 2017-01-26
Make ServiceProvider class public


couven92:
I am writing an extension to the DI system to support Sub-ServiceProviders that are accessible through a Sub-ServiceManager by a specified key (or name).

However, to be able to do that I need to implement a SubServiceProvider class which ideally should work exactly like the normal ServiceProvider, just with the default fallback to ask the root level provider to realize the server if the Sub-ServiceProvider cannot do it.

However, since the ServiceProvider class is internal I cannot derive from it and thus not easily share and extend the feature of the standard DI Provider.

Having the ability for such a nested DI-container could be very beneficial. And although I realize that there are other DI frameworks that support this, I'd really love to build this for the MS DI-system.

If there are big issues with making the ServiceProvider a public class, I am willing to submit a complete PR with support for a Sub-ServiceProvider (and a respective Builder and Manager class) here in this repo.

davidfowl:
We don't have plans to add child container support. There's another long thread on that.

> However, since the ServiceProvider class is internal I cannot derive from it and thus not easily share and extend the feature of the standard DI Provider.

Can you show what exactly you plan to do if the ServiceProvider was public? I don't quite understand it from the description.

couven92:
Example for a use case of a SubServiceProvider:
```csharp
public class Startup {
    public void ConfigureServices(IServiceCollection services) {
        services.AddSubServiceManager((ISubServicesManagerBuilder<string> subServiceManager) => {
            subServiceManager.AddSubServices("subServicesName", (IServiceCollection subServices) => {
                subServices.AddSingleton(new MyMagicService());
            });
        });
    }
    public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory) {
        var serviceProvider = app.ApplicationServices;
        var subServiceManager = serviceProvider.GetRequiredService<ISubServiceManager<string>>();
        IServiceProvider subServiceProvider = subServiceManager.GetServiceProvider("subServicesName");
        var myMagicServiceInstance = subServiceProvider.GetRequiredService<MyMagicService>();
    }
}
```

This way, you could make branches in the DI-container structure, and separate each branch from each other... This makes the following possible:
- Having multiple Database Providers in a DI container
- Having DbContexts with different sources (as in migration applications that move data from one DB to another)
- This would also allow for hosting Web Applications with small modules within the same application scope, where each module could be its own independant application

[...]

PR #477, opened 2017-01-27
Make ServiceProvider class public

couven92:
Ref. Issue #476

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection/DefaultServiceProvider.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/DefaultServiceProvider.cs
@@ -15,22 +15,22 @@ namespace Microsoft.Extensions.DependencyInjection
     /// <summary>
     /// The default IServiceProvider.
     /// </summary>
-    public class ServiceProvider : IServiceProvider, IDisposable
+    public class DefaultServiceProvider : IServiceProvider, IDisposable
     {
         private readonly CallSiteValidator _callSiteValidator;
         private readonly ServiceTable _table;
         private bool _disposeCalled;
         private List<IDisposable> _transientDisposables;
 
-        internal ServiceProvider Root { get; }
+        internal DefaultServiceProvider Root { get; }
         internal Dictionary<object, object> ResolvedServices { get; } = new Dictionary<object, object>();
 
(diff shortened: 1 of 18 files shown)


davidfowl:
I feel like we should rename this class if we make it public. `ServiceProvider` just seems too.. I dunno generic.

/cc @Eilon @pakrym @halter73

couven92:
Some of the collections often inherit from *CollectionName* **Base**

So maybe: `ServiceProviderBase`?

davidfowl:
It's not a base class, it's a default implementation of `IServiceProvider`

couven92:
`ServiceRealizer`? `DIServiceProvider`? `DefaultServiceProvider` (ref `DefaultServiceProviderFactory``)?

I suck at naming, so I do not really have much better suggestions...

cwe1ss:
Would be a lot cooler if you had product names for these libraries! Now you have to use `MicrosoftExtensionsDependencyInjectionServiceProvider`!  😀

pakrym:
I think `DefaultServiceProvider` should be fine

halter73:
I question the wisdom of making ServiceProvider public in the first place. It doesn't serve much of a purpose.

It's not as if using it as a base class achieves anything considering it's only ever exposed as an IServiceProvider and IDisposable. Not coincidentally, GetService and Dispose are the only public members on the class. There are no protected members. This means anything you could achieve with inheritance, you could achieve about just as easily with encapsulation.

Making GetService virtual isn't nearly as useful as it might seem at first glance. The GetService method isn't used for transitive/nested/recursive/what-ever-you-call-them services. That of course makes the default fallback to a root level provider as suggested in #476 difficult to implement in any useful way. At least not in any way that couldn't be achieved just as well with encapsulation. And that's not even getting into the issues related to scoping.

If we do decide to make ServiceProvider public, I would like to see the actual code that needs to reference it.

davidfowl:
@halter73 agreed. See https://github.com/aspnet/DependencyInjection/issues/476#issuecomment-275557577

Eilon:
Let's do another DI triage this coming week. I'll set it up.

couven92:
I have renamed the `ServiceProvider` to `DefaultServiceProvider` as @pakrym suggested in https://github.com/aspnet/DependencyInjection/pull/477#issuecomment-275570866

davidfowl:
Closing this out until we get more clarity on if this is needed

PR #479, opened 2017-02-01
Update to VS 2017

pranavkm:


build/common.props
```diff
@@ -0,0 +1,25 @@
+<Project ToolsVersion="15.0">
+
+  <Import Project="..\version.props" />
+
+  <PropertyGroup>
+    <Product>Microsoft .NET Extensions</Product>
+    <RepositoryUrl>https://github.com/aspnet/Razor</RepositoryUrl>
```
natemcmaster: Fix URL

DependencyInjection.sln
```diff
@@ -6,20 +6,19 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{D26F6F80-63E
 EndProject
```
natemcmaster: Also update lines with Visual Studio 14 => 15

build/common.props
```diff
@@ -0,0 +1,25 @@
...
+    <NetCoreAppImplicitPackageVersion>1.2.0-*</NetCoreAppImplicitPackageVersion>
+    <NetStandardImplicitPackageVersion>1.6.2-*</NetStandardImplicitPackageVersion>
+    <VersionSuffix Condition="'$(VersionSuffix)'!='' AND '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Internal.AspNetCore.Sdk" Version="1.0.1-*" PrivateAssets="All" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETFramework' AND '$(OutputType)'=='library'">
```
natemcmaster: This group can be removed because we never produce packages here for .NET Framework

natemcmaster (changes requested):
Also, add `os: Visual Studio 2017 RC` to appveyor.yml. Just noticed net451 tests aren't running with the Windows-only net451. Will fix that soon.

pranavkm:
🆙 :date

natemcmaster (approved):

natemcmaster:
(Assuming you're aware of the compile issue)

```
ServiceCollectionContainerBuilderExtensions.cs(16,66): error CS0246: The type or namespace name 'IServiceCollection' could not be found (are you missing a using directive or an assembly reference?) 
```

natemcmaster:
Also, just curious...doesn't affect approval.... Why was ActiviatorUtilities.cs renamed to DIActivatorUtilities?

pranavkm:
We get a file name collision with the file imported as a content item from https://github.com/aspnet/Common/blob/dev/shared/Microsoft.Extensions.ActivatorUtilities.Sources/ActivatorUtilities.cs. Wasn't aware of the buidl error - I'll check.

natemcmaster:
> We get a file name collision with the file imported as a content item from 

That shouldn't happen. Probably a bug on NuGet, but could be dotnet/sdk and the way they do default items.

pranavkm:
Ok. I'll try and get a repro for it.

PR #482, opened 2017-02-11
Added ServiceProviderOptions type that flows to the container

davidfowl:
- We'll need to add more options soon
(like disabling transient disposable capture), this change adds an overload to
BuildServiceProvider(), adds a ctor to the DefaultServiceProviderFactory and adds
ctor to ServiceProvider that take the ServiceProviderOptions.

/cc @halter73 @pakrym

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection/DefaultServiceProviderFactory.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/DefaultServiceProviderFactory.cs
@@ -4,6 +4,23 @@ namespace Microsoft.Extensions.DependencyInjection
 {
     public class DefaultServiceProviderFactory : IServiceProviderFactory<IServiceCollection>
     {
+        private readonly ServiceProviderOptions _options = ServiceProviderOptions.Default;
+
+        public DefaultServiceProviderFactory()
+        {
+
+        }
+
+        public DefaultServiceProviderFactory(ServiceProviderOptions options)
+        {
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            _options = options;
+        }
+
         public IServiceCollection CreateBuilder(IServiceCollection services)
         {
             return services;
@@ -11,7 +28,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
         public IServiceProvider CreateServiceProvider(IServiceCollection containerBuilder)
         {
-            return containerBuilder.BuildServiceProvider();
+            return containerBuilder.BuildServiceProvider(_options);
         }
     }
 }
--- a/src/Microsoft.Extensions.DependencyInjection/DefaultServiceProviderFactory.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/DefaultServiceProviderFactory.cs
@@ -4,9 +4,9 @@ namespace Microsoft.Extensions.DependencyInjection
 {
     public class DefaultServiceProviderFactory : IServiceProviderFactory<IServiceCollection>
     {
-        private readonly ServiceProviderOptions _options = ServiceProviderOptions.Default;
+        private readonly ServiceProviderOptions _options;
 
-        public DefaultServiceProviderFactory()
+        public DefaultServiceProviderFactory() : this(ServiceProviderOptions.Default)
         {
 
         }
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
         public static IServiceProvider BuildServiceProvider(this IServiceCollection services)
         {
-            return BuildServiceProvider(services, validateScopes: false);
+            return BuildServiceProvider(services, ServiceProviderOptions.Default);
         }
 
         /// <summary>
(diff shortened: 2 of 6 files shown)


src/Microsoft.Extensions.DependencyInjection/DefaultServiceProviderFactory.cs
```diff
@@ -4,14 +4,31 @@ namespace Microsoft.Extensions.DependencyInjection
...
     public class DefaultServiceProviderFactory : IServiceProviderFactory<IServiceCollection>
     {
+        private readonly ServiceProviderOptions _options = ServiceProviderOptions.Default;
+
+        public DefaultServiceProviderFactory()
+        {
+
+        }
+
+        public DefaultServiceProviderFactory(ServiceProviderOptions options)
```
Meir017: any reason not to merge the two ctor's into one
````csharp
private readonly ServiceProviderOptions _options

public DefaultServiceProviderFactory(ServiceProviderOptions options = null)
{
    _options = options ?? ServiceProviderOptions.Default;
}
````

Comment (davidfowl) [inline src/Microsoft.Extensions.DependencyInjection/DefaultServiceProviderFactory.cs]: We don't use optional parameters in public APIs. Versioning them requires thought (it's doable but we don't use them).

File: src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs [hunk 2]
```diff
@@ -29,7 +29,31 @@ public static IServiceProvider BuildServiceProvider(this IServiceCollection serv
...
+        /// <summary>
+        /// Creates an <see cref="IServiceProvider"/> containing services from the provided <see cref="IServiceCollection"/>
+        /// optionaly enabling scope validation.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/> containing service descriptors.</param>
+        /// <param name="options">
+        /// Configures various service provider behaviors.
+        /// </param>
+        /// <returns>The<see cref="IServiceProvider"/>.</returns>
+        public static IServiceProvider BuildServiceProvider(this IServiceCollection services, ServiceProviderOptions options)
```
davidfowl: Maybe an `Action<ServiceProviderOptions>` overload might be good as well?

```
services.BuildServiceProvider(options =>  { options.ValidateBlah=true; });
```

pakrym:
@davidfowl did you consider adding `UseDefaultServiceProvider(this IWebHostBuilder, options)` overload?

src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
```diff
@@ -29,7 +29,31 @@ public static IServiceProvider BuildServiceProvider(this IServiceCollection serv
         /// <returns>The<see cref="IServiceProvider"/>.</returns>
         public static IServiceProvider BuildServiceProvider(this IServiceCollection services, bool validateScopes)
```
halter73: Obsolete?

src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
halter73: I don't see why that's better than

```
services.BuildServiceProvider(new ServiceProviderOptions { ValidateBlah = true });
```

unless we plan on adding extension methods.

src/Microsoft.Extensions.DependencyInjection/DefaultServiceProviderFactory.cs
```diff
@@ -4,14 +4,31 @@ namespace Microsoft.Extensions.DependencyInjection
 {
     public class DefaultServiceProviderFactory : IServiceProviderFactory<IServiceCollection>
     {
+        private readonly ServiceProviderOptions _options = ServiceProviderOptions.Default;
+
+        public DefaultServiceProviderFactory()
+        {
+
+        }
```
halter73: nit, is slightly better in the event any other logic gets added to the ctor.

```c#
private readonly ServiceProviderOptions _options;

public DefaultServiceProviderFactory()
    : this(ServiceProviderOptions.Default)
{
}
```

src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
davidfowl: Have we obsoleted other API?

src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
davidfowl: Don't need to new up anything that's the only reason we use that pattern. It's also to avoid having to know the namespace the options type (but that's moot here)

davidfowl:
@pakrym thats next (in the hosting repo)

src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
pakrym: I like action overload better, we almost never newup options classes.

davidfowl: I'll leave it out for now.

src/Microsoft.Extensions.DependencyInjection/ServiceCollectionContainerBuilderExtensions.cs
halter73: Yes. https://github.com/aspnet/KestrelHttpServer/blob/796a408be79cb0ddecee8dbde0519f0c5cbc5b6c/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs#L50

davidfowl: https://github.com/aspnet/DependencyInjection/issues/488

PR #489, opened 2017-02-14
Better codegen

davidfowl:
- Elide calls to CaptureDisposable for transient services when the
implementation type isn't disposable
- Elide calls to Convert when the type is already assignable from
the target type

Examples:

```C#
private class ServiceD
{
    public ServiceD(IEnumerable<ServiceA> services)
    {

    }
}

private class ServiceA
{
}

private class ServiceB
{
    public ServiceB(ServiceA serviceA)
    {
        ServiceA = serviceA;
    }

    public ServiceA ServiceA { get; set; }
}

private class ServiceC
{
    public ServiceC(ServiceB serviceB)
    {
        ServiceB = serviceB;
    }

    public ServiceB ServiceB { get; set; }
}
```

Scenario 1:

```C#
GetService<ServiceD>();
```

Before

```C#
public static object GenerateCode(ServiceProvider serviceProvider)
{
	Program.ServiceA[] expr_06 = new Program.ServiceA[1];
	int arg_1A_1 = 0;
	object obj = new Program.ServiceA();
	expr_06[arg_1A_1] = (Program.ServiceA)serviceProvider.CaptureDisposable(obj);
	object obj2 = expr_06;
	object obj3 = new Program.ServiceD((IEnumerable<Program.ServiceA>)serviceProvider.CaptureDisposable(obj2));
	return serviceProvider.CaptureDisposable(obj3);
}
```

After

```C#
public static object GenerateCode(ServiceProvider serviceProvider)
{
	return new Program.ServiceD(new Program.ServiceA[]
	{
		new Program.ServiceA()
	});
}
```

Scenario 2:

```C#
GetService<ServiceC>()
```

Before

```C#
public static object GenerateCode(ServiceProvider serviceProvider)
{
	object obj = new Program.ServiceA();
	object obj2 = new Program.ServiceB((Program.ServiceA)serviceProvider.CaptureDisposable(obj));
	object obj3 = new Program.ServiceC((Program.ServiceB)serviceProvider.CaptureDisposable(obj2));
	return serviceProvider.CaptureDisposable(obj3);
}
```

After

```C#
public static object GenerateCode(ServiceProvider serviceProvider)
{
	return new Program.ServiceC(new Program.ServiceB(new Program.ServiceA()));
}

```

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
@@ -138,15 +138,24 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
             return Expression.NewArrayInit(
                 callSite.ItemType,
                 callSite.ServiceCallSites.Select(cs =>
-                    Expression.Convert(
+                    Convert(
                         VisitCallSite(cs, provider),
                         callSite.ItemType)));
         }
 
         protected override Expression VisitTransient(TransientCallSite callSite, ParameterExpression provider)
         {
+            var implType = callSite.Service.ImplementationType;
+
+            // Elide calls to GetCaptureDisposable if the implemenation type isn't disposable
+            if (implType != null &&
+                !typeof(IDisposable).GetTypeInfo().IsAssignableFrom(implType.GetTypeInfo()))
+            {
+                return VisitCallSite(callSite.ServiceCallSite, provider);
+            }
+
             return Expression.Invoke(GetCaptureDisposable(provider),
-                VisitCallSite(callSite.Service, provider));
+                VisitCallSite(callSite.ServiceCallSite, provider));
         }
 
         protected override Expression VisitConstructor(ConstructorCallSite callSite, ParameterExpression provider)
@@ -155,7 +164,18 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
             return Expression.New(
                 callSite.ConstructorInfo,
                 callSite.ParameterCallSites.Select((c, index) =>
-                        Expression.Convert(VisitCallSite(c, provider), parameters[index].ParameterType)));
+                        Convert(VisitCallSite(c, provider), parameters[index].ParameterType)));
+        }
+
+        private static Expression Convert(Expression expression, Type type)
+        {
+            // Don't convert if the expression is already assignable
+            if (type.GetTypeInfo().IsAssignableFrom(expression.Type.GetTypeInfo()))
+            {
+                return expression;
+            }
+
+            return Expression.Convert(expression, type);
         }
 
         protected override Expression VisitScoped(ScopedCallSite callSite, ParameterExpression provider)
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/ClosedIEnumerableService.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/ClosedIEnumerableService.cs
@@ -3,13 +3,11 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 
 namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
 {
     internal class ClosedIEnumerableService : IService
     {
-        private readonly Type _itemType;
         private readonly ServiceEntry _serviceEntry;
 
         public ClosedIEnumerableService(Type itemType, ServiceEntry entry)
@@ -39,7 +37,7 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
                 list.Add(provider.GetResolveCallSite(service, callSiteChain));
                 service = service.Next;
             }
-            return new ClosedIEnumerableCallSite(_itemType, list.ToArray());
+            return new ClosedIEnumerableCallSite(ServiceType, list.ToArray());
         }
     }
 }
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/ClosedIEnumerableService.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/ClosedIEnumerableService.cs
@@ -14,9 +14,9 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
 
         public ClosedIEnumerableService(Type itemType, ServiceEntry entry)
         {
-            _itemType = itemType;
+            ServiceType = itemType;
             _serviceEntry = entry;
-            ImplementationType = _itemType.MakeArrayType();
+            ImplementationType = itemType.MakeArrayType();
         }
 
         public IService Next { get; set; }
@@ -26,7 +26,7 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
             get { return ServiceLifetime.Transient; }
         }
 
-        public Type ServiceType => _itemType;
+        public Type ServiceType { get; }
 
         public Type ImplementationType { get; }
 
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/ClosedIEnumerableService.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/ClosedIEnumerableService.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
         {
             _itemType = itemType;
             _serviceEntry = entry;
+            ImplementationType = _itemType.MakeArrayType();
         }
 
         public IService Next { get; set; }
@@ -26,6 +27,8 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
         }
 
         public Type ServiceType => _itemType;
+
+        public Type ImplementationType { get; }
 
         public IServiceCallSite CreateCallSite(ServiceProvider provider, ISet<Type> callSiteChain)
         {
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/FactoryService.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/FactoryService.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
 
         public Type ServiceType => Descriptor.ServiceType;
 
-        public Type ImplementationType => Descriptor.ImplementationType;
+        public Type ImplementationType => null;
 
         public IServiceCallSite CreateCallSite(ServiceProvider provider, ISet<Type> callSiteChain)
         {
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/FactoryService.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/FactoryService.cs
@@ -24,6 +24,8 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
 
         public Type ServiceType => Descriptor.ServiceType;
 
+        public Type ImplementationType => Descriptor.ImplementationType;
+
         public IServiceCallSite CreateCallSite(ServiceProvider provider, ISet<Type> callSiteChain)
         {
             return this;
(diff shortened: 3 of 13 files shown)


src/Microsoft.Extensions.DependencyInjection/ServiceLookup/ClosedIEnumerableService.cs
```diff
@@ -27,6 +28,8 @@ public ServiceLifetime Lifetime
 
         public Type ServiceType => _itemType;
```
pakrym: Remove field, use only property.

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
```diff
@@ -186,7 +186,7 @@ public void Dispose()
             }
         }
 
-        internal object CaptureDisposable(object service)
+        internal virtual object CaptureDisposable(object service)
```
davidfowl: Maybe I should revert this virtual call?

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/FactoryService.cs
```diff
@@ -24,6 +24,8 @@ public ServiceLifetime Lifetime
 
         public Type ServiceType => Descriptor.ServiceType;
 
+        public Type ImplementationType => Descriptor.ImplementationType;
```
halter73: I think this line is confusing since `Descriptor.ImplementationType` is always null. Reading this in isolation, I would think that wasn't the case.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/InstanceService.cs
```diff
@@ -27,6 +27,8 @@ public ServiceLifetime Lifetime
 
         public Type ServiceType => Descriptor.ServiceType;
 
+        public Type ImplementationType => Descriptor.ImplementationType;
```
halter73: Same confusing property since ImplementationType is always null. This should probably be Descriptor.ImplementationInstance.GetType().

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/ClosedIEnumerableService.cs
```diff
@@ -3,19 +3,18 @@
...
     {
-        private readonly Type _itemType;
         private readonly ServiceEntry _serviceEntry;
 
         public ClosedIEnumerableService(Type itemType, ServiceEntry entry)
         {
-            _itemType = itemType;
+            ServiceType = itemType;
             _serviceEntry = entry;
+            ImplementationType = itemType.MakeArrayType();
```
halter73: This happens to be an array today, but that's really an implementation detail. It could be changed to any IEnumerable implementation. Since this is only used to see if IDisposable is implemented, why not just make it null like we do for factories?

davidfowl: So we don't generate `provider.CaptureDisposable(new T[] { ... })` when generating the code for `IEnumerable<T>`. See the PR description for an example (it's the first one).

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
halter73: Probably. I'm not a huge fan of unit testing non-functional changes.

halter73: To clarify, it would be worth testing if you could mock Dispose in fake services in your tests, but since dispose isn't called either way, at worst we regress perf. We should have real perf tests instead of testing a call (which you just made virtual ironically) gets elided.

davidfowl: Having performance tests is fine but I think having a way to verify the call isn't actually made is just as valuable (if not more).

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/ClosedIEnumerableService.cs
halter73: Makes sense. It just seems likely that the ImplementationType could become out of sync with the real implementation type if we optimize this later using a different IEnumerable implementation.

This is especially true considering the call to create the array is in CallSiteExpressionBuilder/RuntimeResolver and not in this class. I guess this lack of locality is a downside of the visitor pattern in general.

No matter what IEnumerable implementation we use, it's unlikely to be disposable, but who knows what else we might use IService.ServiceType for.

halter73: IService.ImplementationType*

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
halter73: I could be convinced of the value of the test, but arguing it's more valuable then real benchmarks that are of course monitored for regressions is laughable. I understand we live in the real world though, and what I'm asking for could be a long time coming.

At the very least we should verify that this change doesn't move perf in the wrong direction by making this virtual. You are eliding the very (short) method you just made virtual. If in practice, this method still gets called far more often than it gets elided, the entire change might not be worth it if that means virtualizing the method.

Do we know if expression tree compilation will ever inline this method? Could that be changed by visualizing it? Even if CaptureDisposable was never inlined in the first place, doesn't mean making this virtual is free.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/ClosedIEnumerableService.cs
davidfowl: > Makes sense. It just seems likely that the ImplementationType could become out of sync with the real implementation type if we optimize this later using a different IEnumerable implementation.

Yes, it can but we code review things and the thing that would break is not disposing the new ienumerable implementation (assuming thats the only correctness issue) and we'd likely have tests for that.

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
davidfowl: > I could be convinced of the value of the test, but arguing it's more valuable then real benchmarks that are of course monitored for regressions is laughable. I understand we live in the real world though, and what I'm asking for could be a long time coming.

@pakrym is going to add some benchmark.net tests to this repo.

> Do we know if expression tree compilation will ever inline this method? Could that be changed by visualizing it? Even if CaptureDisposable was never inlined in the first place, doesn't mean making this virtual is free.

I'm going to change it from being virtual so we don't even theoretically affect the performance. Will add an internal callback instead.

halter73: By that standard, this PR theoretically affects performance 😄

davidfowl: It's definitely better.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/InstanceService.cs
```diff
@@ -16,6 +16,7 @@ internal class InstanceService : IService, IServiceCallSite
         public InstanceService(ServiceDescriptor descriptor)
         {
             Descriptor = descriptor;
+            ImplementationType = descriptor.ImplementationInstance?.GetType();
```
halter73: Can ImplementationInstance ever actually be null here?

davidfowl: Nope, it's null checked before it gets here. Will remove.

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
```diff
@@ -188,6 +191,8 @@ public void Dispose()
 
         internal object CaptureDisposable(object service)
         {
+            _captureDisposableCallback?.Invoke(service);
```
halter73: Is this better than virtual dispatch even when _captureDisposableCallback is null?

davidfowl: I didn't look at the ASM 😉 . Now we're down in the weeds. Branching vs virtual dispatch. It *does* clean up the test code since we don't need a derived service provider (that was kinda gross).

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
benaadams: This whole project is still pretty exotic to me; but can give it an investigation. Also I worry about the hours your team is keeping 😄 

Is `disposable == null` a common path? If so the lock portion could be factored out to a separate function which might make it more likely to inline.

davidfowl: That's still worth doing as this code path gets used for singletons. Another PR 😄

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
pakrym: Just make `_transientDisposables` internal and inspect it.

davidfowl: That's not enough. It's about calls to this method not captured disposables. In the cases above, the transient services aren't disposable.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/InstanceService.cs
pakrym: Why not use `descriptor.GetImplementationType()` ?

davidfowl: Didn't even see that but it doesn't matter here anyways.

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
pakrym: Maybe it's cleaner to inspect generated expression trees?

davidfowl: That's in no way cleaner than this. It's just harder to do.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/ClosedIEnumerableService.cs
pakrym: If this data is used only for expression tree  generation can we just make a function in `CallSiteExpressionBuilder` that does this?

davidfowl: I can do that in another PR if you are worried about the overhead of calling a mostly null delegate in a method that'll be elided 😄 , there's enough churn here already and this adds nothing.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
```diff
@@ -138,15 +138,24 @@ protected override Expression VisitClosedIEnumerable(ClosedIEnumerableCallSite c
...
+                    Convert(
                         VisitCallSite(cs, provider),
                         callSite.ItemType)));
         }
 
         protected override Expression VisitTransient(TransientCallSite callSite, ParameterExpression provider)
         {
+            var implType = callSite.Service.ImplementationType;
+
+            // Elide calls to GetCaptureDisposable if the implemenation type isn't disposable
```
pranavkm: ~~implemenation~~ implementation

src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
halter73: Inspecting the expressions definitely produces cleaner product code. You just don't want to write more complicated tests. The test is super fragile to begin with because it's testing an implementation detail that isn't at all observable given using black box testing. Maybe the perf difference is observable, but we haven't even established that yet.

davidfowl: Correct and I'm not going to 😄

Issue #490, opened 2017-02-14
Get better on this benchmark


davidfowl:
There's a nice DI container comparison here:

https://github.com/danielpalme/IocPerformance

We should get higher on this list 😄

benaadams:
Would IL.Emit be too far? (may or may not be relevant: https://blogs.msdn.microsoft.com/seteplia/2017/02/01/dissecting-the-new-constraint-in-c-a-perfect-example-of-a-leaky-abstraction/)

seesharper:
Optimizing service lookup is probably one area to look into. It's not necessarily a problem with the generated code, but how the factory delegates are looked up during service resolution. The Dictionary and/or ConcurrentDictionary are good all-purpose structures, but there are faster alternatives such as AVL trees.

dadhi:
@benaadams,

IL.Emit is not too far, here is [FastExpressionCompiler](https://github.com/dadhi/FastExpressionCompiler) which I use in DryIoc, which is higher in this benchmark.

@seesharper, and faster lookup AVL trees - yes. [Here](https://github.com/dadhi/ImTools) we go :-)

aspnet-hello:
This issue was moved to aspnet/Home#2343

PR #502, opened 2017-03-11
Added benchmarks project and basic benchmark 

davidfowl:
- Added a benchmark project (using benchmark.NET)
- Copied most of the infrastructure from kestrel and tweaked it a little 
- Added a resolve benchmark

|    Method |        Mean |    StdErr |     StdDev | Scaled | Scaled-StdDev |            RPS |  Gen 0 | Allocated |
|---------- |------------ |---------- |----------- |------- |-------------- |--------------- |------- |---------- |
|      NoDI |   6.9489 ns | 0.1808 ns |  0.9905 ns |   1.00 |          0.00 | 143,906,789.32 | 0.0112 |      24 B |
| Transient |  51.2131 ns | 0.3905 ns |  2.1389 ns |   7.47 |          0.82 |  19,526,255.52 | 0.0091 |      24 B |
|    Scoped | 106.3770 ns | 4.2441 ns | 23.2460 ns |  15.53 |          3.71 |   9,400,532.68 |      - |       0 B |
| Singleton | 123.1784 ns | 3.6458 ns | 19.9689 ns |  17.98 |          3.42 |   8,118,307.04 |      - |       0 B |


khellang:
Does RPS bring any value here?

davidfowl:
@khellang That's a good question. I like those big numbers 😄

benaadams:
> Does RPS bring any value here?

Its easier to evaluate than `0.0001 ns`

benaadams:
Quite fast... might need to stick it in a loop to be accurate :)

pakrym:
I don't like copying same code around, would prefer having package with benchmark.net specific code

davidfowl:
> I don't like copying same code around, would prefer having package with benchmark.net specific code

I knew you were going to say that. Will postpone this until then.

pakrym:
`Program.cs` code can be shared too

davidfowl:
Like 5% of the code, the enum (which is a hack) isn't sharable. It really needs to be based on type names, not an enum.

pakrym:
We need to make it type based not enum based anyway, so it would be 100%

davidfowl:
It needs to be rewritten is my point. Anyways, I'll add a package to Common with those 2 shared functions.

davidfowl:
Also filed https://github.com/dotnet/BenchmarkDotNet/issues/386

arespr:
here is nice comparison:

http://www.palmmedia.de/blog/2011/8/30/ioc-container-benchmark-performance-comparison

davidfowl:
@arespr yes, see https://github.com/aspnet/DependencyInjection/issues/490

pakrym:
@davidfowl do we want to contribute RPS column to BDN?

davidfowl:
@pakrym Yes, I was just asking @benaadams if he wrote it. It really belongs there.

benaadams:
RPS came from @mattwarren

PR #503, opened 2017-03-12
Added scope validation benchmark

davidfowl:
- Renamed benchmarks to *Benchmark

```
                       Method |        Mean |    StdErr |     StdDev |        Op/s | Scaled | Scaled-StdDev |  Gen 0 | Allocated |
----------------------------- |------------ |---------- |----------- |------------ |------- |-------------- |------- |---------- |
                    Transient |  43.5473 ns | 0.4130 ns |  2.2620 ns | 22963524.25 |   1.00 |          0.00 | 0.0095 |      24 B |
 TransientWithScopeValidation |  68.7545 ns | 1.3960 ns |  7.6460 ns | 14544498.22 |   1.58 |          0.19 | 0.0076 |      24 B |
    ScopedWithScopeValidation | 105.1968 ns | 1.9990 ns | 10.9491 ns |   9505991.5 |   2.42 |          0.28 |      - |       0 B |
 SingletonWithScopeValidation | 136.7783 ns | 2.0167 ns | 11.0459 ns |  7311100.08 |   3.15 |          0.29 |      - |       0 B |
```

### DIFF
--- a/test/Microsoft.Extensions.DependencyInjection.Performance/GetServiceBenchmark.cs
+++ b/test/Microsoft.Extensions.DependencyInjection.Performance/GetServiceBenchmark.cs
@@ -7,7 +7,7 @@ using BenchmarkDotNet.Attributes;
 namespace Microsoft.Extensions.DependencyInjection.Performance
 {
     [Config(typeof(CoreConfig))]
-    public class ResolvePerformance
+    public class GetServiceBenchmark
     {
         private const int OperationsPerInvoke = 50000;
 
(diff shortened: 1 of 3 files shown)


pakrym:
Open generic case?

davidfowl:
As part of scope validation? Does it affect the code path differently? I ask because I didn't have an open generic for the basic `GetServiceBenchmark` either.

pakrym:
Nope, my bad. Scope validation adds two overheads:

1. During any service creation it walks call site tree - so we need a benchmark with longer callsite tree that causes it to walk more.
2. During service resolution it does a lookup - benchmark that just does transient resolution (because it's fastest and slowdown is easiest to see).

PR #505, opened 2017-03-15
Dispose services in reverse creation order

pakrym:
@davidfowl

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/DependencyInjectionSpecificationTests.cs
+++ b/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/DependencyInjectionSpecificationTests.cs
@@ -662,10 +662,10 @@ namespace Microsoft.Extensions.DependencyInjection.Specification
             // Arrange
             var serviceCollection = new TestServiceCollection();
             serviceCollection.AddSingleton<FakeDisposeCallback>();
-            serviceCollection.AddSingleton<IFakeOuterService, FakeDisposableCallbackOuterService>();
+            serviceCollection.AddTransient<IFakeOuterService, FakeDisposableCallbackOuterService>();
             serviceCollection.AddSingleton<IFakeMultipleService, FakeDisposableCallbackInnerService>();
-            serviceCollection.AddSingleton<IFakeMultipleService, FakeDisposableCallbackInnerService>();
-            serviceCollection.AddSingleton<IFakeMultipleService, FakeDisposableCallbackInnerService>();
+            serviceCollection.AddScoped<IFakeMultipleService, FakeDisposableCallbackInnerService>();
+            serviceCollection.AddTransient<IFakeMultipleService, FakeDisposableCallbackInnerService>();
             serviceCollection.AddSingleton<IFakeService, FakeDisposableCallbackInnerService>();
             var serviceProvider = CreateServiceProvider(serviceCollection);
 
--- a/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/DependencyInjectionSpecificationTests.cs
+++ b/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/DependencyInjectionSpecificationTests.cs
@@ -655,5 +655,30 @@ namespace Microsoft.Extensions.DependencyInjection.Specification
             Assert.Same(expected.MultipleService, actual.MultipleService);
             Assert.Same(expected.ScopedService, actual.ScopedService);
         }
+
+        [Fact]
+        public void DisposesInReverseOrderOfCreation()
+        {
+            // Arrange
+            var serviceCollection = new TestServiceCollection();
+            serviceCollection.AddSingleton<FakeDisposeCallback>();
+            serviceCollection.AddSingleton<IFakeOuterService, FakeDisposableCallbackOuterService>();
+            serviceCollection.AddSingleton<IFakeMultipleService, FakeDisposableCallbackInnerService>();
+            serviceCollection.AddSingleton<IFakeMultipleService, FakeDisposableCallbackInnerService>();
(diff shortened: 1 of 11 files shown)


src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
```diff
@@ -164,34 +164,42 @@ protected override Expression VisitScoped(ScopedCallSite callSite, ParameterExpr
...
+                resolvedVariable, VisitCallSite(callSite.ServiceCallSite, provider));
 
             var addValueExpression = Expression.Call(
                 resolvedServices,
                 AddMethodInfo,
                 keyExpression,
-                resolvedExpression);
+                resolvedVariable);
+
+            Expression captureDisposableExpression = Expression.Invoke(
```
davidfowl: Errr?

halter73: We never tested the actual performance in terms of op/s, just memory allocations. If you look closely at your [original PR](https://github.com/aspnet/DependencyInjection/pull/291) you'll notice that the memory allocations measured are mostly related ServiceProvider.CreateScope()/ctor, and there are no measurements related to the actual capturing disposables of any kind.

If you're really concerned memory you could store some sort of ordering metadata with the captured services (a la Autofac) or pool the Dictionary/List like @benaadams [suggested in the original PR](https://github.com/aspnet/DependencyInjection/pull/291#issuecomment-140784285). Both sound kinda complicated and might even hurt perf in some scenarios.

It might be a good idea to add a BenchmarkDotNet scenario that covers capturing disposables.

src/Microsoft.Extensions.DependencyInjection.Specification.Tests/DependencyInjectionSpecificationTests.cs
```diff
@@ -655,5 +655,30 @@ public static TheoryData ServiceContainerPicksConstructorWithLongestMatchesData
...
+        public void DisposesInReverseOrderOfCreation()
+        {
+            // Arrange
+            var serviceCollection = new TestServiceCollection();
+            serviceCollection.AddSingleton<FakeDisposeCallback>();
+            serviceCollection.AddSingleton<IFakeOuterService, FakeDisposableCallbackOuterService>();
+            serviceCollection.AddSingleton<IFakeMultipleService, FakeDisposableCallbackInnerService>();
+            serviceCollection.AddSingleton<IFakeMultipleService, FakeDisposableCallbackInnerService>();
+            serviceCollection.AddSingleton<IFakeMultipleService, FakeDisposableCallbackInnerService>();
+            serviceCollection.AddSingleton<IFakeService, FakeDisposableCallbackInnerService>();
```
halter73: Nit: might make sense to have a test that mixes service lifetimes. Could be relevant if we go back to tracking disposables using multiple collections.

src/Microsoft.Extensions.DependencyInjection.Specification.Tests/Fakes/FakeService.cs
```diff
@@ -2,9 +2,61 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 
 namespace Microsoft.Extensions.DependencyInjection.Specification.Fakes
 {
+    public class FakeDisposeCallback
```
halter73: There's already 1000 .cs files in Specification.Tests/Fakes, just put all the public classes in their own file.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
```diff
@@ -164,34 +164,42 @@ protected override Expression VisitScoped(ScopedCallSite callSite, ParameterExpr
...
+                    resolvedVariable
                 },
                 Expression.IfThen(
                     Expression.Not(tryGetValueExpression),
-                    Expression.Block(assignExpression, addValueExpression)),
-                resolvedExpression);
+                    Expression.Block(
+                        assignExpression,
+                        addValueExpression,
+                        captureDisposableExpression)),
```
halter73: Suuuper nit: match ordering with runtime resolver. I know it doesn't really matter.

halter73 (changes requested):
A benchmark would be nice to alleviate @davidfowl's concerns.

src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs
davidfowl: Yes, we need measurements. Also `ConcurrentBag` did get way less allocatey in .NET Core 2.0 so maybe it's worth a remeasure.

pakrym: This change would only add scoped and singletons to that list, there aren't many singletons and scope objects are short-lived so I am not super worried about this.

davidfowl: I still want measurements.

davidfowl: Also, all of this work to elide calls to CaptureDisposable this should be repeated here https://github.com/aspnet/DependencyInjection/blob/dev/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/CallSiteExpressionBuilder.cs#L150.

davidfowl (approved):

Issue #360, opened 2016-02-12
Add ability to have multiple service types per definition


davidfowl:
I want to be able to get the same instance for this type of definition:

``` C#
public class Bar : IBar, IBaz
{
}
```

``` C#
services.AddSingleton<IBar, Bar>();
services.AddSingleton<IBaz, Bar>();
var sp = services.BuildServiceProvider();
var i1 = sp.GetService<IBar>();
var i2 = sp.GetService<IBaz>();
Assert.Same(i1, i2); // Fails
```

This works but is less efficient.

``` C#
services.AddSingleton<Bar>();
services.AddSingleton<IBar>(sp => sp.GetService<Bar>());
services.AddSingleton<IBaz>(sp => sp.GetService<Bar>());
var i2 = sp.GetService<IBaz>();
var i1 = sp.GetService<IBar>();
var i2 = sp.GetService<IBaz>();
Assert.Same(i1, i2); // Passes
```

This also works but I have to know the entire dependency graph to construct bar:

``` C#
var bar = new Bar(); // Hope Bar has no services
services.AddSingleton<IBar>(bar);;
services.AddSingleton<IBaz>(bar);
var i2 = sp.GetService<IBaz>();
var i1 = sp.GetService<IBar>();
var i2 = sp.GetService<IBaz>();
Assert.Same(i1, i2); // Passes
```

We'd have to look at how to support this, should it be a  service descriptor?

```
services.AddSingleton<IBar, IBaz, Bar>();
```

We also have to see if other DI containers support this.

dadhi:
[DryIoc](https://bitbucket.org/dadhi/dryioc/overview) supports this via [RegisterMany](https://bitbucket.org/dadhi/dryioc/wiki/RegisterResolve#markdown-header-registermany) and more specifically via [RegisterMapping](https://bitbucket.org/dadhi/dryioc/wiki/RegisterResolve#markdown-header-registermapping).

seesharper:
With regards to the API I think it is important to specify all interfaces along with the implementing type (or factory delegate) in one single registration method. 

Consider this scenario 

```
services.AddSingleton<IBar>(provider => new Bar());
services.AddSingleton<IBaz>(provider => new Bar());
```

In this scenario we have no way to look at the implementing type at registration time to determine if both interfaces should resolve to the same instance or not. The intent here could just as well be that we want two singletons, one for IBar and one for IBaz.

[...]

PR #509, opened 2017-03-24
Different singleton registrations resolve to same instance

yufeih:
- Singleton uses same instance when registered with different service types
- Scoped instances are same within a scope but are different across scopes
- Singleton instances are disposed exactly once

https://github.com/aspnet/DependencyInjection/issues/360


davidfowl:
This is a breaking change. If we decide to support this, I'm not sure we can just break the existing behavior.

yufeih:
I see, like `services.AddSingleton<IBar, IBaz, Bar>();`

Eilon:
Closing per https://github.com/aspnet/DependencyInjection/issues/360#issuecomment-289603154

Issue #338, opened 2015-12-16
How to dispose top-level IServiceProvider


khellang:
Since `IServiceProvider` doesn't implement `IDisposable`, and `ServiceProvider` (which _does_ implement it) is internal, how are you supposed to dispose a "top-level service provider", i.e. the result of calling `IServiceCollection.BuildServiceProvider()`?

I could of course do a `(provider as IDisposable)?.Dispose()`, but it would be nicer if the interface implemented `IDisposable`. That way it would work better with generic constraints etc.

I understand that there's little to do with the interface itself, since it's an old interface that's been reused, but would it be possible to add a separate one?

Something along the lines of

``` csharp
namespace Microsoft.Extensions.DependencyInjection
{
    public interface IServiceProvider : System.IServiceProvider, IDisposable { }
}
```

I _think_ I've gotten around the problem by immediately creating a scope, which is disposable, and use its provider, but I don't think it's a good solution.

What was the reasoning for using the existing interface? And why is everything internal? :stuck_out_tongue_closed_eyes:

Eilon:
Moving this to Backlog as we are not planning to make any changes for this at this time.

khellang:
What about an answer and close it instead?

davidfowl:
https://github.com/aspnet/Mvc/issues/3726#issuecomment-163535556

Nope :smile:

khellang:
I don't get it. Is that the answer? How does that relate to this?

I don't really care about the interface, but there has to be some way to get a hold of the concrete type to call dispose on it, without resorting to casting etc.?

Eilon:
Yeah not sure what @davidfowl is talking about there...

But the answer for now is that unfortunately we don't have a good answer on this. I think creating a scope immediately is an acceptable workaround in the meantime. This is certainly something we can reconsider.

davidfowl:
We'd have to expose the concrete type (which we were trying to avoid) from BuildServiceProvider

davidfowl:
@Eilon we *can* consider this for 2.0. Make a breaking change to the return value of BuildServiceProvider and make the internal `ServiceProvider` public.

Eilon:
Let's do another fun DI triage 😄

PR #510, opened 2017-03-28
Make ServiceProvider public

davidfowl:
- Make ServiceProvider public so that it can be returned
from BuildServiceProvider().
- Make it sealed since it can't really be extended.
- Changed tests to stop down casting.

#338

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Linq.Expressions;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
@@ -15,7 +14,7 @@ namespace Microsoft.Extensions.DependencyInjection
     /// <summary>
     /// The default IServiceProvider.
     /// </summary>
-    internal class ServiceProvider : IServiceProvider, IDisposable
+    public sealed class ServiceProvider : IServiceProvider, IDisposable
     {
         private readonly CallSiteValidator _callSiteValidator;
         private readonly ServiceTable _table;
@@ -33,7 +32,7 @@ namespace Microsoft.Extensions.DependencyInjection
         // CallSiteRuntimeResolver is stateless so can be shared between all instances
         private static readonly CallSiteRuntimeResolver _callSiteRuntimeResolver = new CallSiteRuntimeResolver();
 
-        public ServiceProvider(IEnumerable<ServiceDescriptor> serviceDescriptors, ServiceProviderOptions options)
+        internal ServiceProvider(IEnumerable<ServiceDescriptor> serviceDescriptors, ServiceProviderOptions options)
         {
             Root = this;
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Extensions.DependencyInjection/ServiceProvider.cs
```diff
@@ -15,7 +14,7 @@ namespace Microsoft.Extensions.DependencyInjection
     /// <summary>
     /// The default IServiceProvider.
     /// </summary>
-    internal class ServiceProvider : IServiceProvider, IDisposable
+    public sealed class ServiceProvider : IServiceProvider, IDisposable
```
davidfowl: I know it's evil but @benaadams will be happy

benaadams: Aren't any virtual calls; might as well make it non-sealed? (Then people can extend/enhance in a non-interfering way)

davidfowl: The other reason I made it sealed is to avoid things like https://github.com/aspnet/DependencyInjection/issues/476

pakrym: And people can always wrap.

benaadams: Can't add extra data to the type and then use an `as` cast to surface it with a wrap...

src/Microsoft.Extensions.DependencyInjection/exceptions.netcore.json
```diff
@@ -0,0 +1,9 @@
+[
+  {
+    "OldTypeId": "public static class Microsoft.Extensions.DependencyInjection.ServiceCollectionContainerBuilderExtensions",
+    "OldMemberId": "public static System.IServiceProvider BuildServiceProvider(this Microsoft.Extensions.DependencyInjection.IServiceCollection services)",
+    "NewTypeId": "public static class Microsoft.Extensions.DependencyInjection.ServiceCollectionContainerBuilderExtensions",
+    "NewMemberId": "public static Microsoft.Extensions.DependencyInjection.ServiceProvider BuildServiceProvider(this Microsoft.Extensions.DependencyInjection.IServiceCollection services)",
+    "Kind": "Modification"
```
dougbu: FYI API Check doesn't actually care much about additions to regular classes and it has a bug when handling `Modification` exceptions in `interface`s (aspnet/BuildTools#154). Might as well make this a `Removal` and leave out the `New*` properties.

davidfowl: When the tool spits out what it wants me to write I'll do that. It took me way too long to figure this out.

khellang:
👍

halter73:
@davidfowl Will hosting continue to call [(_applicationServices as IDisposable)?.Dispose();](https://github.com/aspnet/Hosting/blob/2df3639535c085f60e186d39b39bfa5bb25ea9b4/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L278), or will it only dispose `ServiceProvider`s?

davidfowl:
@halter73 I'll fix that once it propagates.

davidfowl:
Should have added Release to the https://github.com/aspnet/DependencyInjection/blob/dev/src/Microsoft.Extensions.DependencyInjection.Abstractions/IServiceProviderFactory.cs

PR #522, opened 2017-04-29
improve: fast return for unclosed type.

Cologler:


davidfowl:
@Cologler Thanks for the PR.
- Can you add more details to the PR summarizing what you are fixing. 
- Can you also write tests for those fixes.

Cologler:
the PR was created for a wrong example.

Issue #521, opened 2017-04-29
[feature request] support  complex generic type


Cologler:
easy to reproduce:

``` cs
class Program
{
    static void Main(string[] args)
    {
        var sc = new ServiceCollection();
        sc.AddSingleton(typeof(ListConverter<>).GetInterfaces().Single(), typeof(ListConverter<>));
        var provider = sc.BuildServiceProvider(); // System.ArgumentException throw here.
        var service = provider.GetService(typeof(ListConverter<int>));
        Console.Read();
    }

    public interface IValueConverter<out TResult> { }

    public class ListConverter<T> : IValueConverter<List<T>> { }
}
```

davidfowl:
What's the scenario you're after? What are you trying to do?

Cologler:
I am trying to use dependency injection build a type convert system.

so I can call:

``` cs
provider.GetService<IValueConverter<Stream>>();
provider.GetService<IValueConverter<List<Stream>>>();
provider.GetService<IValueConverter<HashSet<Stream>>>();
provider.GetService<IValueConverter<Dictionary<string, List<Stream>>>>();
```

seesharper:
This is partially closed generics which is a rather complicated area. 
I really don't know if this is something the default implementation should support as it makes it a lot harder to write an adapter.

PR #523, opened 2017-04-30
implement partially closed generics service.

Cologler:
- implement partially closed generics service.
- add unittest.

now I can use:

``` cs
public interface ITestInterface1<TI1, TI2, TI3> { }
public class TestClass2<TC1> : ITestInterface1<TC1, List<TC1>, Dictionary<string, HashSet<TC1>>> { }

var serviceCollection = new ServiceCollection();
serviceCollection.AddTransient(
    typeof(TestClass1<,>).GetTypeInfo().GetInterfaces().Single(),
    typeof(TestClass1<,>));
var provider = serviceCollection.BuildServiceProvider();
provider.GetRequiredService<ITestInterface1<string, List<string>, Dictionary<string, HashSet<string>>>>();
```

see #521.

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/PartiallyClosedGenericsService.cs
+++ b/src/Microsoft.Extensions.DependencyInjection/ServiceLookup/PartiallyClosedGenericsService.cs
@@ -1,0 +1,256 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Reflection;
+using System.Linq;
+
+namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
+{
+    internal class PartiallyClosedGenericsService : IGenericService
+    {
+        private readonly ServiceDescriptor _descriptor;
+        private readonly GenericTypeTree _tree;
+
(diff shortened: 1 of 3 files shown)


Cologler:
TODO: my englist is poor, need someone help me format the exception message.

davidfowl:
@Cologler we don't add features to this container that aren't supported in other containers (for many reasons) because that makes it harder to write an adapter. Can you do some research to see if this is feature is supported in some other popular containers?

If not, we won't take this PR.

Cologler:
this feature will not change any of the old behavior (old behavior always throw a ArgumentException).

why that make it harder to write an adapter?

davidfowl:
> this feature will not change any of the old behavior (old behavior always throw a ArgumentException).

Changing old behavior isn't the problem.

This container serves as a minimal container implementation that supports the most common subset   of DI features. Features we add that don't exist in other containers make it harder to both a container adapter. You can see the existing container adapters here:

https://github.com/aspnet/DependencyInjection#using-other-containers-with-microsoftextensionsdependencyinjection

And the existing specification tests:

https://github.com/aspnet/DependencyInjection/tree/dev/src/Microsoft.Extensions.DependencyInjection.Specification.Tests

So, once again, we won't take this if it doesn't work on those existing containers.

Cologler:
Oh, I am already understood. If other developer try to use this feature, then they cannot change DI container.

So, if I need to use this feature, I should create my own branch.

PR #526, opened 2017-05-11
Update test SDK and xunit version

natemcmaster:
Update to Microsoft.NET.Test.Sdk 15.3.0 and xunit 2.3.0-beta2.

Cleanup a duplicate test (thanks to better xunit diagnostics in xunit 2.3 :) )

cref https://github.com/aspnet/Coherence-Signed/issues/49

cc @Eilon - OSS approval for new xunit version

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -6,7 +6,7 @@
     <InternalAspNetCoreSdkVersion>2.1.0-*</InternalAspNetCoreSdkVersion>
     <MoqVersion>4.7.1</MoqVersion>
     <NETStandardImplicitPackageVersion>$(BundledNETStandardPackageVersion)</NETStandardImplicitPackageVersion>
-    <TestSdkVersion>15.0.0</TestSdkVersion>
-    <XunitVersion>2.2.0</XunitVersion>
+    <TestSdkVersion>15.3.0-*</TestSdkVersion>
+    <XunitVersion>2.3.0-beta2-*</XunitVersion>
   </PropertyGroup>
 </Project>
--- a/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj
+++ b/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj
@@ -19,8 +19,4 @@
     <PackageReference Include="xunit.extensibility.core" Version="$(XunitVersion)" />
   </ItemGroup>
 
-  <ItemGroup>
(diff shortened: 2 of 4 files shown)


build/dependencies.props
```diff
@@ -6,7 +6,7 @@
     <InternalAspNetCoreSdkVersion>2.1.0-*</InternalAspNetCoreSdkVersion>
     <MoqVersion>4.7.1</MoqVersion>
     <NETStandardImplicitPackageVersion>$(BundledNETStandardPackageVersion)</NETStandardImplicitPackageVersion>
-    <TestSdkVersion>15.0.0</TestSdkVersion>
-    <XunitVersion>2.2.0</XunitVersion>
+    <TestSdkVersion>15.3.0-*</TestSdkVersion>
```
Eilon: Should these use exact versions? Or is it safe to let them float? I kind of prefer exact versions because it makes sure no one will pull a switcheroo on us and then our build breaks.

(Same goes for xUnit version.)

natemcmaster: I think it's safe to float. Here's my thinking:

Floating xunit: the next update to NuGet would be `2.3.0-beta3-*` so we won't get automatically lifted off beta2. Also, I used floating this a matter of convenience because the full version of `xunit` and `xunit.runner.visualstudio` differ slightly by their build numbers.

Floating Microsoft.NET.Test.Sdk to `15.3.0-*` should also be ok. We mirror Microsoft.NET.Test.Sdk in our gatekeeper build, so the version on [aspnetcore-ci-dev](https://dotnet.myget.org/feed/aspnetcore-ci-dev/package/nuget/Microsoft.NET.Test.Sdk) is ahead of [the version](https://www.nuget.org/packages/Microsoft.net.test.sdk) on nuget.org. As new nightlies roll in, our gatekeeper build should catch issues before they make it to nuget.org. (hopefully)

src/Microsoft.Extensions.DependencyInjection.Specification.Tests/Microsoft.Extensions.DependencyInjection.Specification.Tests.csproj
```diff
@@ -19,8 +19,4 @@
     <PackageReference Include="xunit.extensibility.core" Version="$(XunitVersion)" />
   </ItemGroup>
 
-  <ItemGroup>
-    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
```
natemcmaster: FYI - should no longer be added automatically when using VS 15.3.0 preview. :)

build/dependencies.props
Eilon: Hmm I guess that's ok!

PR #527, opened 2017-05-16
Add IDesignTimeServiceProviderFactory

### LINKED ISSUE #524: Interface to formalize how tooling obtains the application's service provider
This a more general purpose alternative to the feature described in https://github.com/aspnet/EntityFramework/issues/8331. 

Currently tools that require access to the application's runtime services and configuration (e.g. EF Core tools, ASP.NET Core Scaffolding, possibly Razor tooling and others) need to hardcode heuristics which 
- often require for tools to take additional unwanted dependencies, e.g. EF tools brings ASP.NET Core Hosting in 1.0 even if it can be used in non-ASP.NET Core applications
- only work if the application follows specific (often version-specific) conventions to build the service provider (e.g. `Startup.ConfigureServices()` in ASP.NET Core 1.0, `Program.BuildWebHost()` in ASP.NET Core 2.0, etc.) 
- and require restoring to bespoke fallback mechanisms (like `IDbContextFactory` in EF) if the convention is not matched

This issue is about defining a common accessor interface that tooling can scan for at design time which provides access to the application's service provider. We believe this approach:  

- Could be used in ASP.NET Core applications as well as any other application that uses our DI
- Could be leveraged by any tool that requires access to the application's services, rather than being component-specific
- Removes the need of hardcoding heuristics in tools by moving the necessary code to the application
- Enables project templates can incorporate an implementation of the interface that is tailored to the patterns they used to construct the application's service provider, e.g. for the ASP.NET Core 2.0 templates the implementation can return `Program.BuildWebHost(args).Services`. If the mechanism changes in the future, the template can be updated
- Addresses all the concerns described in https://github.com/aspnet/EntityFramework/issues/8331, e.g. it allows EF Core tooling to discover any `DbContext` registered with `AddDbContext` regardless of their location, making `IDbContextFactory` unnecessary for most common cases

Example:

``` C#
public class DesignTimeServiceProviderAccessor : IDesignTimeServiceProviderAccessor
{
    public IServiceProvider GetServiceProvider(string[] args)
        => Program.BuildWebHost(args).Services;
}
```

Open issues:
- Still need to run the idea by the architect :smile:
- General shape and naming of the interface (or abstract type?) is up in the air
- Need to decide where we want to put it in templates

cc @bricelam @ajcvickers @DamianEdwards @davidfowl

bricelam:
Provides access to the application's IServiceProvider to design-time services.

Resolves #524

cc @ajcvickers @DamianEdwards @davidfowl @divega @Eilon

### DIFF
--- a/src/Microsoft.Extensions.DependencyInjection.Abstractions/IDesignTimeServiceProviderFactory.cs
+++ b/src/Microsoft.Extensions.DependencyInjection.Abstractions/IDesignTimeServiceProviderFactory.cs
@@ -1,0 +1,22 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// A factory for creating the application's <see cref="IServiceProvider"/> at design time. Implement this interface
+    /// to enable design-time services that need access to the application's services. Design-time services will
+    /// automatically discover implementations of this interface.
+    /// </summary>
+    public interface IDesignTimeServiceProviderFactory
+    {
+        /// <summary>
+        /// Creates a new instance of the application's <see cref="IServiceProvider"/>.
+        /// </summary>
+        /// <param name="args">Arguments provided by the design-time service.</param>
+        /// <returns>The application's <see cref="IServiceProvider"/>.</returns>
+        IServiceProvider CreateServiceProvider(string[] args);
+    }
+}


src/Microsoft.Extensions.DependencyInjection.Abstractions/IDesignTimeServiceProviderFactory.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// A factory for creating the application's <see cref="IServiceProvider"/> at design time. Implement this interface
+    /// to enable design-time services that need access to the application's services. Design-time services will
+    /// automatically discover implementations of this interface.
```
ajcvickers: automatically discover implementations of this interface _in the startup assembly_?

ajcvickers:
@Eilon @DamianEdwards Who else do you want to look at/sign-off on this before we merge?

src/Microsoft.Extensions.DependencyInjection.Abstractions/IDesignTimeServiceProviderFactory.cs
bricelam: It's really up to the tool... I debated whether to say "in the project's assembly"

ajcvickers: Fair enough. Where will EF tools look for it? Can we envision other tools looking elsewhere?

bricelam: maybe "application assembly?"

ericwj: I understood you were talking about formalizing the Startup class pattern. This is much more limited. XDesProc will need another interface with the same shape to obtain the runtime IServiceProvider. Navigation uses the page type exclusively [Frame.Navigate](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.inavigate) and the system instantiates the page with that (and an optional object that is meant to be passed as-is to the page, defined on Frame). And well that part is off course shared with C++. Not sure whether that is definitively prohibitive for what I was thinking this could become.

ericwj: It now sais I reviewed? That's bold

divega: Is there something clearer we can say instead of "design-time services" maybe just "tools" or "tooling"?

src/Microsoft.Extensions.DependencyInjection.Abstractions/IDesignTimeServiceProviderFactory.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// </summary>
+    public interface IDesignTimeServiceProviderFactory
```
davidfowl: Can we change the name to not have design time anything?

davidfowl: It's really just a factory.

src/Microsoft.Extensions.DependencyInjection.Abstractions/IDesignTimeServiceProviderFactory.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// Creates a new instance of the application's <see cref="IServiceProvider"/>.
+        /// </summary>
+        /// <param name="args">Arguments provided by the design-time service.</param>
+        /// <returns>The application's <see cref="IServiceProvider"/>.</returns>
+        IServiceProvider CreateServiceProvider(string[] args);
```
davidfowl: The args is pretty strange. Does it need to be on the interface? Can we flow context in the constructor of the implementation

src/Microsoft.Extensions.DependencyInjection.Abstractions/IDesignTimeServiceProviderFactory.cs
divega: It is a factory with a specific purpose. Since tooling will be looking for implementations of this interface for that specific purpose it would be really problematic if arbitrary implementations used the same interface. 

We could create an interface that is just a general purpose factory and have this one derive from it, but not sure what the scenario is.

Eilon: Yeah I do think this interface is specifically for design-time. If we think there's a broader use-case, I'd be interested. I saw the comment about wanting this for runtime, but I don't think this is the solution you'd want for that anyway.

ericwj: If the name is just a tag, tools could discover an attribute perhaps?

ericwj: @Eilon don't think it is needed at runtime. For as far as I've thought it through so far, the discovery would be done by XDesProc only if you open a XAML file, to instantiate views and their dependencies at design time without any bootstrapping having done already, allowing constructor arguments to be resolved. There is also overhead in finding it, which is especially important for apps and not strictly necessary since there is code that definitely runs before navigation.

ericwj: The scenario that a common non-generic interface would formalize is to delegate the factory somewhere else altogether. For say Prism that'd allow better separation of concerns in PrismApplication on .NET Standard, in which `IServiceProviderFactory<TContainerBuilder>` is only half the solution to e.g. light up Autofac on top of DependencyInjection.Abstractions. Today I'm using [this pattern](http://autofac.readthedocs.io/en/latest/integration/aspnetcore.html) which means so far I had taken dependency on DependencyInjection just to use ServiceCollection. Not sure that kind of scenario warrants defining an interface, but if it exists, I'd prefer to use it.

davidfowl: There's nothing here that makes this specific to design time except for the name. Do we really think this thing is that specific? Also the args makes it looks like it was designed for console applications. I think we should remove them because it's beginning to look like isn't that general purpose... 

Maybe we should call it, `IShouldBeCalledFromTheCommandLineWithArgsServiceProviderFactory`

divega: Interfaces have the semantics we assign to them :smile_cat: What makes `IServiceScope` something more than a way to get a service provider? Is it really just the fact that it implements `IDisposable` or does the name and our shared knowledge of what it is for that make it special?

I am fine with passing the args in the constructor unless there are technical issues with it. I actually had the same thoughts a few days ago and I already forgot why I abandoned them :smile:

davidfowl: > Interfaces have the semantics we assign to them 😸 What makes IServiceScope something more than a way to get a service provider?

It's disposable. Looking back if we were doing this over again, we would have a single interface represent the scope and the top level container.

divega: Yes, it is `IDisposable` (I said that already). It also has a property for the service provider.  I just don't think that is enough to define what the essence of what a service scope is. Name and shared understanding of its role contribute quite a bit.

All that said, I would be fine with using a general purpose service provider factory interface + an attribute if there is where you want to go.

ericwj: I'm thinking about how calls into Startup gets their arguments. Basically that's DI too. Though last time I checked it was a lot of code to make it look that way and not extensible. Making that extensible would be a good reason to have both design time something that can be resolved by tools and a discoverable common one that is always used. Such a pattern *could* also use discovery at runtime, but it could again probably just be Main that brings it before creating the host builder with an API call to replace the default or just a call on host builder itself. But going after this will be way bigger a change than envisioned I guess.

src/Microsoft.Extensions.DependencyInjection.Abstractions/IDesignTimeServiceProviderFactory.cs
bricelam: I considered that too. I like that it becomes optional, but hate how undiscovered it is.

bricelam: Plus the implementation would get uglier.
```C#
class DesignTimeServices : IDesignTimeServiceProviderFactory
{
    private readonly string[] _args;

    public DesignTimeServices(string[] args)
    {
        _args = args;
    }

    public virtual IServiceProvider CreateServiceProvider()
    {
        return Program.BuildWebHost(_args).Services;
    }
}
```

davidfowl:
I changed my mind, this should go into another package. Lets call it `Microsoft.Extensions.DesignTimeSupport`

Eilon:
I'm ok with something like `Microsoft.Extensions.DesignTimeSupport`.

ajcvickers:
@Eilon @davidfowl After discussing with @divega and @bricelam yesterday I think we are going to back off from this for now. It's hard to fight @davidfowl on something he doesn't like very much, and @bricelam has just done the work to allow EF to call BuildWebHost directly, so we don't need this for 2.0. I still think it is a good thing to do, but we can always do it later when there isn't a crunch on for other things. I was waiting to confirm with @divega on this, but he is out today.

bricelam:
Recalling until we've sorted out the design.

divega:
As per @ajcvickers's previous comment, we are punting this. Not that we no longer think it is a good idea, even beyond ASP.NET Core, but it apparent that cooking a good design we can all agree on is going to take longer and we don't have the time now.

Re the name of a package that could host this interface in the future, the .NET idiomatic thing for "design time support" is `.Design`. That said, it is not clear to me that `Microsoft.Extensions` would convey the right scope of applicability for this interface.

davidfowl:
> Re the name of a package that could host this interface in the future, the .NET idiomatic thing for "design time support" is .Design. That said, it is not clear to me that Microsoft.Extensions would convey the right scope of applicability for this interface.

I don't have a problem with the idea of this interface but putting it in this package really doesn't fit. The interface is very specific. The other benefit of another assembly is that all of the support code for activating these things can also be added there.

PR #534, opened 2017-06-07
Target .NET Standard 2.0

natemcmaster:
Part of aspnet/Home#2045.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -2,10 +2,9 @@
   <PropertyGroup>
     <AspNetCoreVersion>2.0.0-*</AspNetCoreVersion>
     <BenchmarkDotNetVersion>0.10.3</BenchmarkDotNetVersion>
-    <CoreFxVersion>4.3.0</CoreFxVersion>
     <InternalAspNetCoreSdkVersion>2.1.0-*</InternalAspNetCoreSdkVersion>
     <MoqVersion>4.7.1</MoqVersion>
-    <NETStandardImplicitPackageVersion>$(BundledNETStandardPackageVersion)</NETStandardImplicitPackageVersion>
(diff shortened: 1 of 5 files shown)


natemcmaster:
⌚️ on https://github.com/aspnet/EventNotification/pull/70 to go through first.

build/dependencies.props
```diff
@@ -2,10 +2,9 @@
   <PropertyGroup>
     <AspNetCoreVersion>2.0.0-*</AspNetCoreVersion>
     <BenchmarkDotNetVersion>0.10.3</BenchmarkDotNetVersion>
-    <CoreFxVersion>4.3.0</CoreFxVersion>
     <InternalAspNetCoreSdkVersion>2.1.0-*</InternalAspNetCoreSdkVersion>
     <MoqVersion>4.7.1</MoqVersion>
-    <NETStandardImplicitPackageVersion>$(BundledNETStandardPackageVersion)</NETStandardImplicitPackageVersion>
```
henkmollema: Just a question here, is `NETStandardImplicitPackageVersion` not necessary anymore? I see some differences in the `dependencies.props` files accross repos (for example here https://github.com/aspnet/HtmlAbstractions/pull/39/files).

pranavkm: @henkmollema This gets set to the bundled version by the SDK: https://github.com/dotnet/sdk/blob/a4bceb67b160c72318a581d87d52796e4fa31794/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Sdk.DefaultItems.targets#L55 if you're targeting `netstandard2.0`. One of the HtmlAbstractions continues to target ns1.0 which is why it needs to be explicitly specified.

henkmollema: Thanks @pranavkm, makes sense!

PR #539, opened 2017-06-30
Add RemoveAll extension method to service collection

pakrym:


pranavkm:
What do we need this for?

pakrym:
Removing all logger providers if you used default webhost builder.

src/Microsoft.Extensions.DependencyInjection.Abstractions/Extensions/ServiceCollectionDescriptorExtensions.cs
```diff
@@ -540,5 +540,38 @@ public static void TryAddSingleton<TService>(this IServiceCollection collection,
...
+        /// <param name="serviceType">The service type to remove.</param>
+        /// <returns></returns>
+        public static IServiceCollection RemoveAll(this IServiceCollection collection, Type serviceType)
+        {
+            if (serviceType == null)
+            {
+                throw new ArgumentNullException(nameof(serviceType));
+            }
+
+            foreach (var serviceDescriptor in collection
```
pranavkm: Isn't `t `IServiceCollection` a list type? Can't you remove by index?

test/Microsoft.Extensions.DependencyInjection.Tests/ServiceCollectionServiceExtensionsTest.cs
```diff
@@ -395,5 +395,23 @@ public void Replace_ReplacesFirstServiceWithMatchingServiceType()
...
+            // Arrange
+            var collection = new ServiceCollection
+            {
+                new ServiceDescriptor(typeof(IFakeService), typeof(FakeService), ServiceLifetime.Transient),
+                new ServiceDescriptor(typeof(IFakeService), typeof(FakeService), ServiceLifetime.Transient),
+                new ServiceDescriptor(typeof(IFakeService), typeof(FakeService), ServiceLifetime.Transient)
+            };
+
+            // Act
+            collection.RemoveAll<IFakeService>();
```
pranavkm: Can you keep some times in the list (so we know it's not just removing everything)?

src/Microsoft.Extensions.DependencyInjection.Abstractions/Extensions/ServiceCollectionDescriptorExtensions.cs
pakrym: Indexes change when you start removing things, or it would need to go in reverse order. I found this one simplest to understand.

pranavkm (approved):

PR #542, opened 2017-07-04
Fix ImplementationType of ConstructorCallSite and FactoryCallSite

### LINKED ISSUE #541: Transient and scoped services are not disposed when resolved using a compiled call site
This breaks EF DbContext pooling and could lead to a memory leak.

AndriySvyryd:
Fixes #541


pakrym:
Thank you for fixing this. I changed the fix a bit to address situation where service type is not disposible but implementation is.

PR #561, opened 2017-07-26
Shorten file names

ryanbrandenburg:
We're shortening the file names here to avoid running into max path problems.


src/DI.Specification.Tests/DI.Specification.Tests.csproj
```diff
@@ -5,6 +5,8 @@
   <PropertyGroup>
     <Description>Suite of xUnit.net tests to check for container compatibility with Microsoft.Extensions.DependencyInjection.</Description>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <AssemblyName>Microsoft.Extensions.DependencyInjection.Specification.Tests</AssemblyName>
+    <RootNamespace>Microsoft.Extensions.DependencyInjection.Specification</RootNamespace>
```
Eilon: Any reason to have this deviate? Normally if these two settings aren't set, their default value is the project's filename or something, no?

test/DI.Tests/DI.Tests.csproj
```diff
@@ -5,11 +5,13 @@
   <PropertyGroup>
     <TargetFrameworks>netcoreapp2.0;net461</TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netcoreapp2.0</TargetFrameworks>
+    <AssemblyName>Microsoft.Extensions.DependencyInjection.Tests</AssemblyName>
+    <RootNamespace>Microsoft.Extensions.DependencyInjection</RootNamespace>
```
Eilon: See previous comment.

src/DI.Specification.Tests/DI.Specification.Tests.csproj
ryanbrandenburg: Don't we not want namespaces in test repos to include `*.Tests`? Or is that not what this property means?

ryanbrandenburg: Oh, this is in `src` not `test`. Still, all the namespaces in this project are `Microsoft.Extensions.DependencyInjection.Specification` not `Microsoft.Extensions.DependencyInjection.Specification.Tests`.

ryanbrandenburg:
@Eilon and I talked through the RootNamespace stuff offline. We decided that it should be safe to do it this way provided there are no RESX files in the project, and so long as we check that the resultant DLL's are the same.

For posterity: The benefit of having different AssemblyName and RootNamespace is that when someone does "Add new *.cs" they get a file with the correct namespace (no `.Tests`) by default.

pakrym:
Do we really want to call it DI? Isn't `DependencyInjection` enough?

natemcmaster:
DI is too short. Would it be okay to keep the folders like this?
```
DependencyInjection
DependencyInjection.Abstractions
DependencyInjection.Specification.Tests
```

Eilon:
Why is DI too short? The container folder is already `DependencyInjection/src/`. We're needing to save every possible character here...

pakrym:
Just feels like overacronymization

Eilon:
If you have alternative ideas on how to solve this, that'd be great. But we need to do as much as we can to shorten names.

natemcmaster:
The only alternative idea I have is what I suggested above. My concern is consistency. What will the pattern be on other repos? Should we be using acronyms as we do this? e.g. does "Microsoft.Extensions.Configuration/" become "CFG/"?

Also, @ryanbrandenburg are you doing this across all repos?

Eilon:
`Microsoft.Extensions.Configuration` could easily become `Config`. I think that's both very readable and understandable.

PR #562, opened 2017-07-27
Move ActivatorUtilities to DI

pakrym:
https://github.com/aspnet/Common/issues/245#event-1180973596


pranavkm:
@Tratcher \ @davidfowl  could we modify `Http.Abstractions` to reference `DI.Abstractions`? DI's copy of `ActivatorUtils` is exposed via DI.Abstractions and `Http.Abstractions` is the only project that references the sources version of `ActivatorUtils` today.

davidfowl:
@pranavkm nah. Let's not take a new dependency

shared/Microsoft.Extensions.ActivatorUtilities.Sources/ActivatorUtilities.cs
```diff
@@ -0,0 +1,305 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Runtime.ExceptionServices;
+
+namespace Microsoft.Extensions.Internal
```
pranavkm: Could we use `ifdefs` to replace https://github.com/aspnet/DependencyInjection/blob/dev/src/DI.Abstractions/DIActivatorUtilities.cs? Something like https://github.com/aspnet/Common/blob/dev/shared/Microsoft.Extensions.WebEncoders.Sources/WebEncoders.cs#L18?

shared/Microsoft.Extensions.ActivatorUtilities.Sources/ActivatorUtilities.cs
```diff
@@ -0,0 +1,305 @@
...
+{
+    internal delegate object ObjectFactory(IServiceProvider serviceProvider, object[] arguments);
+
+    // Do not take a dependency on this class unless you are explicitly trying to avoid taking a
```
pranavkm: This comment is not super useful. I don't think anyone actually reads these messages.

shared/Microsoft.Extensions.ActivatorUtilities.Sources/sharedsources.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <ItemGroup>
+    <Compile Include="$(NuspecBasePath)\..\Microsoft.Extensions.ParameterDefaultValue.Sources\ParameterDefaultValue.cs">
```
pranavkm: Can a source package reference another source package? Could we potentially leave `ParameterDefaultValue` in common and use a package reference?

pakrym: I tried hard but wasn't able to make it work. My other though was just to leave them slip and reference both where we reference ActivatorUtils

pranavkm: Fair enough

pakrym:
:up::date: `#ifdef`ed ActivatorUtils

pranavkm:
:shipit:

PR #569, opened 2017-08-16
Upgrade to xunit 2.3.0-beta4

natemcmaster:
Also, switch to using Directory.Build.props/targets and add xunit.runner.json


test/DI.Tests/ServiceCollectionServiceExtensionsTest.cs
```diff
@@ -211,11 +211,14 @@ public static TheoryData TryAddImplementationTypeData
...
-            var collection = new ServiceCollection();
-            collection.Add(ServiceDescriptor.Transient(expectedServiceType, expectedServiceType));
+            var collection = new ServiceCollection
+            {
+                ServiceDescriptor.Transient(expectedServiceType, expectedServiceType)
+            };
 
             // Act
             addAction(collection);
+            collection.TryAdd(ServiceDescriptor.Describe(expectedServiceType, expectedImplementationType, expectedLifetime));
```
pakrym: What does this fix?

natemcmaster: xunit1026: "unused theory parameters". The MemberData has items in it that aren't really needed for the test. We could either duplicate part of the memberdata into a colleciont that doesn't have these extra parameters, or use them.

pakrym: I don't like adding this like if accomplishes nothing but disabling a warning, which is false positive in this case.

natemcmaster: I think it fits with the description of the test: "TryAdd_WithType_DoesNotAddDuplicate". But you're welcome to refactor this test if it bothers you enough.

pakrym: `addAction` already calls `TryAdd`. Please remove this line.

PR #571, opened 2017-08-17
Revert test change from #569

pakrym:


test/DI.Tests/ServiceCollectionServiceExtensionsTest.cs
```diff
@@ -207,8 +207,11 @@ public static TheoryData TryAddImplementationTypeData
         public void TryAdd_WithType_DoesNotAddDuplicate(
             Action<IServiceCollection> addAction,
             Type expectedServiceType,
+#pragma warning disable xUnit1026
```
natemcmaster: Add an explanation why this data isn't used.

natemcmaster (approved):

Issue #586, opened 2017-09-19
Improve the circular dependency exception message


AndriySvyryd:
Currently it says `A circular dependency was detected for the service of type 'MyType'`.

It should provide all types in the cycle.

Eilon:
You mean like this?

`A circular dependency was detected for the service of type that has a circular dependency was detected for the service of type that has a circular dependency was detected for the service of type that has a circular dependency was detected for the service of type that has a circular dependency was detected for the service of type that has a circular dependency was detected for the service of type that has a circular dependency was detected for the service of type that has a circular dependency was detected for the service of type that has a circular dependency was detected for the service of type that has a circular dependency.`?

AndriySvyryd:
@Eilon Sounds perfect, as I always say - redundancy leads to clarity.

Eilon:
@pakrym I'm assigning to you to resolve when https://github.com/aspnet/DependencyInjection/pull/592 is merged.

PR #592, opened 2017-10-10
Circular dependency exception improve

AndreyTretyak:
Pull request was inspired by issue #586 .

- add resolution path to exception message of circular dependency, for example:
```
A circular dependency was detected for the service of type 'TypeA'.
Resolution path:
Resolving 'TypeA' by activating 'TypeA'.
Resolving 'TypeB' by activating 'TypeB'.
Resolving 'TypeC' by activating 'TypeC'.
Resolving 'TypeA'.
```
- add unit test for circular dependency in IEnumerable;
- add unit test for circular dependency A -> B -> C -> B;
- add `CallSiteChain` that will allow to display resolution path in other exception.


davidfowl:
The type printing needs some love. I'd like to see something like this:

```
A circular dependency was detected for the service of type 'TypeA'.

TypeA -> B -> C -> D -> TypeA
```

I'd also like to see how you print open generics and closed generics. It might be useful to use our type pretty printer here https://github.com/aspnet/Common/blob/b9bc076c782af4a9664b34b561fd91d37e6edda4/shared/Microsoft.Extensions.TypeNameHelper.Sources/TypeNameHelper.cs

src/DI/Resources.resx
```diff
@@ -150,4 +150,15 @@
   <data name="DirectScopedResolvedFromRootException" xml:space="preserve">
     <value>Cannot resolve {1} service '{0}' from root provider.</value>
   </data>
+  <data name="ResolutionPathSeparator" xml:space="preserve">
+    <value> -&gt; </value>
```
Eilon: This should not be a resource (it's not localizable).

src/DI/Resources.resx
```diff
@@ -150,4 +150,15 @@
...
+  </data>
+  <data name="ResolutionPathServiceAndImplementationType" xml:space="preserve">
+    <value>{0}({1})</value>
```
Eilon: This should not be a resource (it's not localizable).

src/DI/Resources.resx
```diff
@@ -150,4 +150,15 @@
...
+    <comment>{0} = service type, {1} = implementation type</comment>
+  </data>
+  <data name="ResolutionPathServiceType" xml:space="preserve">
+    <value>{0}</value>
```
Eilon: This should not be a resource (it's not localizable).

test/DI.Tests/CircularDependencyTests.cs
```diff
@@ -2,8 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using Microsoft.Extensions.DependencyInjection.Tests.Fakes;
 using Xunit;
+using static Microsoft.Extensions.Internal.TypeNameHelper;
```
Eilon: I would recommend not using any helper here at all and putting the stringified names in the tests, which I think will make the tests easier to understand and verify because you can see what the exception messages will really look like.

src/DI/ServiceLookup/CallSiteChain.cs
```diff
@@ -0,0 +1,85 @@
...
+
+                builder.Append(Resources.ResolutionPathSeparator);
+            }
+
+            builder.AppendFormat(Resources.ResolutionPathServiceType, TypeNameHelper.GetTypeDisplayName(currentlyResolving));
+        }
+
+        private struct ChainItemInfo
+        {
+            public int Order { get; }
```
Eilon: Is Order ever used? I thought the order is always a natural order due to the sequence of calls in the chain.

AndreyTretyak:
Circular dependency error message using new format and `TypeNameHelper`  for type printing looks like this:

- Simple case:
```
A circular dependency was detected for the service of type 'Microsoft.Extensions.DependencyInjection.Tests.Fakes.IndirectCircularDependencyA'.
Microsoft.Extensions.DependencyInjection.Tests.Fakes.IndirectCircularDependencyA -> Microsoft.Extensions.DependencyInjection.Tests.Fakes.IndirectCircularDependencyB -> Microsoft.Extensions.DependencyInjection.Tests.Fakes.IndirectCircularDependencyC -> Microsoft.Extensions.DependencyInjection.Tests.Fakes.IndirectCircularDependencyA
```

- Enumerable:
```
A circular dependency was detected for the service of type 'Microsoft.Extensions.DependencyInjection.Tests.Fakes.SelfCircularDependency'.
System.Collections.Generic.IEnumerable<Microsoft.Extensions.DependencyInjection.Tests.Fakes.SelfCircularDependency> -> Microsoft.Extensions.DependencyInjection.Tests.Fakes.SelfCircularDependency -> Microsoft.Extensions.DependencyInjection.Tests.Fakes.SelfCircularDependency
```

- Different service and implementation type:
```
A circular dependency was detected for the service of type 'Microsoft.Extensions.DependencyInjection.Tests.Fakes.ISelfCircularDependencyWithInterface'.
Microsoft.Extensions.DependencyInjection.Tests.Fakes.SelfCircularDependencyWithInterface -> Microsoft.Extensions.DependencyInjection.Tests.Fakes.ISelfCircularDependencyWithInterface(Microsoft.Extensions.DependencyInjection.Tests.Fakes.SelfCircularDependencyWithInterface) -> Microsoft.Extensions.DependencyInjection.Tests.Fakes.ISelfCircularDependencyWithInterface
```

- Generics:
```
A circular dependency was detected for the service of type 'Microsoft.Extensions.DependencyInjection.Tests.Fakes.SelfCircularDependencyGeneric<string>'.
Microsoft.Extensions.DependencyInjection.Tests.Fakes.SelfCircularDependencyGeneric<int> -> Microsoft.Extensions.DependencyInjection.Tests.Fakes.SelfCircularDependencyGeneric<string> -> Microsoft.Extensions.DependencyInjection.Tests.Fakes.SelfCircularDependencyGeneric<string>
```

src/DI/ServiceLookup/CallSiteFactory.cs
```diff
@@ -118,12 +114,13 @@ private IServiceCallSite TryCreateOpenGeneric(Type serviceType, ISet<Type> callS
...
         }
 
-        private IServiceCallSite TryCreateEnumerable(Type serviceType, ISet<Type> callSiteChain)
+        private IServiceCallSite TryCreateEnumerable(Type serviceType, CallSiteChain callSiteChain)
         {
             if (serviceType.IsConstructedGenericType &&
                 serviceType.GetGenericTypeDefinition() == typeof(IEnumerable<>))
             {
                 var itemType = serviceType.GenericTypeArguments.Single();
+                callSiteChain.Add(serviceType);
```
Eilon: Hmm, this was not in the chain before, but now it is, so more types are in the circular dependency check. Is this correct? I'm worried about a potential breaking change here because now more checks will be happening. (But, maybe the code was always wrong, and now just a better exception will be thrown.)

src/DI/ServiceLookup/CallSiteFactory.cs
```diff
@@ -227,8 +224,10 @@ private IServiceCallSite ApplyLifetime(IServiceCallSite serviceCallSite, object
             }
         }
 
-        private IServiceCallSite CreateConstructorCallSite(Type serviceType, Type implementationType, ISet<Type> callSiteChain)
+        private IServiceCallSite CreateConstructorCallSite(Type serviceType, Type implementationType, CallSiteChain callSiteChain)
         {
+            callSiteChain.Add(serviceType, implementationType);
```
Eilon: See earlier comment about potential breaking change.

src/DI/ServiceLookup/CallSiteChain.cs
AndreyTretyak: Order used during resolution path display `_callSiteChain.OrderBy(p => p.Value.Order)` in `AppendResolutionPath` function.

Although `Dictionary<TKey,TValue>` currently store pair in natural order this behavior not guaranteed.

We could use `System.Collections.Specialized.OrderedDictionary`, but it's not generic and will cause a lot of boxing/unboxing. 

This could be solved if we get rid of `ChainItemInfo` and use `ImplementationType` as Value, but we won't have ability to add some additional info. For example: initially I had enum `ResolutionType`(ConstructorCall/InstanceUse/EnumerableCreate/FactoryRun) that could be used to display each case differently.

src/DI/ServiceLookup/CallSiteFactory.cs
AndreyTretyak: If I understand everything correctly with this changes we will have slightly less checks, because currently type added to chain before we decide how to resolve it (Use Instance / Call Constructor / Create Enumerable / Run Factory). Instance Use and Factory Run can't cause circular reference so I propose add type to chain only in cases of Enumerable Creation and Constructor Call (with implementation type for more detailed error).

pakrym: `IEnumerable<>` it was, it was just added here: https://github.com/aspnet/DependencyInjection/pull/592/files/bac0c91230d57467eb46c2d1ff8da4ee35f7018a#diff-a7622bce2aa4a5b7dbb9e510b8fe8f86L80

Now instead of adding in one common point it gets added in multiple. Mostly because more information is required (implementation type)

Eilon: Got it, thanks for the explanations.

src/DI/ServiceLookup/CallSiteChain.cs
Eilon: Ok, but no one is ever setting Order as far as I can tell, so I think there's still a problem, no?

And yes, you're absolutely right that we can't depend on Dictionary returning the same order. Beyond a certain size, it's nearly guaranteed that it won't return the same order.

AndreyTretyak: Order set in `Add` method, when we create `ChainItemInfo`. It equals to count of items added in chain before, to preserve natural order.

Eilon: Oh sorry I totally missed that, you are correct. Looks good!

src/DI/ServiceLookup/CallSiteChain.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public void CheckCircularDependency(Type serviceType)
+        {
+            if (_callSiteChain.ContainsKey(serviceType))
+            {
+                throw new InvalidOperationException(CreateCircularDependencyExceptionMessage(serviceType));
+            }
+        }
+
+        public void Remove(Type serviceType)
```
halter73: Nit: It doesn't look like this is called anywhere.

pakrym: It's used in https://github.com/aspnet/DependencyInjection/pull/592/files#diff-a7622bce2aa4a5b7dbb9e510b8fe8f86R87

PR #597, opened 2017-10-16
Add RepositoryRoot

ryanbrandenburg:


ryanbrandenburg:
This failure seems unrelated to my change, @pakrym is investigating.

PR #606, opened 2017-10-30
Add tracing to DI

### LINKED ISSUE #605: Add EventSource/EventCounter tracing and metrics for DependencyInjection
The primary goal here is to give users insight as to when services are constructed and how many services they have active in their application. This can be useful when correlated to requests and helps give more insight into memory issues.

There is also a useful effect of putting `EventSource`s in DI. Since DI is below Logging in the dependency graph, it has no logging. However, `EventSource` is in corefx, which means DI can use it.

I'm going to do this one via a PR rather than listing proposed counters since it's relatively simple.

See https://gist.github.com/anurse/af1859663ac91c6cf69c820cebe92303 for some guidance on adding `EventSource`s and `EventCounter`s to ASP.NET projects.

analogrelay:
Resolves #605 

**Cannot be merged until ValueStopwatch.IsActive is available from Common**

I ran the benchmarks before and after the change and saw a <1% change with EventSources disabled. I have some data here: https://gist.github.com/anurse/63ec880c3440998f6cb9e01d7a35eacd

I did experiment with an event for **resolving** a service (i.e. wrapping GetService). It caused a 15% regression in perf, even when disabled. This is somewhat in keeping with my observations about EventSource, which is that they are fast but not completely free and in really hot paths they need to be managed differently. One option I considered if we really care about them is that you could invalidate all the cached call sites when those detailed events are enabled and compile them in, but that seems way out of scope for now :).

A counter I considered but didn't find a clean way to implement in my first pass was "Call Site Depth" which would track how "deep" a call site goes (i.e. how many recursive activations are needed to satisfy a dependency).

Events
* `RealizingService` Level: `Informational` - `ServiceProviderEngine.RealizeService` is about to be called to construct a service. Payload:
  * `serviceType` - Assembly-Qualified Type Name of the requested service
  * `expectedImplementationType` - Assembly-Qualified Type Name of the implementation if known (or `null` if not)
  * `callSiteType` - Assembly-Qualified Type Name of the call site (for lifetime info?). Not sure how useful this is...
* `RealizedService` Level: `Informational` - `ServiceProviderEngine.RealizeService` was just called
  * `serviceType` - Assembly-Qualified Type Name of the requested service
  * `implementationType` - Assembly-Qualified Type Name of the implementation (which is now known for sure)
  * `callSiteType` - Assembly-Qualified Type Name of the call site (for lifetime info?). Not sure how useful this is...
  * `durationInMilliseconds` - Duration of time taken to realize the service. This **includes** the time taken to realize dependent services.
* `ScopeStarted` Level: `Verbose` - A Scope was just started. No payload (there's nothing to provide!)
* `ScopeEnded` Level: `Verbose` - A Scope was just disposed. No payload (there's nothing to provide!)
* `DisposedService` Level: `Verbose` - A service was just disposed. Payload:
  * `implementationType` - Assembly-Qualified Type Name of the service implementation being disposed.

Counters

* `ServiceRealizationDuration` - Records the `durationInMilliseconds` value from `RealizedService`
* `ServicesRealized` - Records a `1.0f` value for every `RealizedService` event


pakrym:
Wrapping get service is hard also because This would not be correct because `GetService` is only called for top level service so no down level dependencies would be tracked this way. To correctly implement `GetService` we would have to both call it from runtime resolver (https://github.com/aspnet/DependencyInjection/blob/dev/src/DI/ServiceLookup/CallSiteRuntimeResolver.cs) and compile calls into expressions (https://github.com/aspnet/DependencyInjection/blob/dev/src/DI/ServiceLookup/CallSiteExpressionBuilder.cs).

anurse:
@pakrym Yeah, that's basically what I came to, a model where we effectively "recompile" all the call sites when the verbose logging is enabled to have them emit the events.

src/DI/ServiceLookup/ServiceProviderEngine.cs
```diff
@@ -51,7 +52,15 @@ protected ServiceProviderEngine(IEnumerable<ServiceDescriptor> serviceDescriptor
...
             {
                 _callback?.OnCreate(callSite);
-                return RealizeService(callSite);
+
+                var timer = DependencyInjectionEventSource.Log.RealizingService(callSite);
+
+                var instance = RealizeService(callSite);
+
+                // REVIEW: NEED TO REPLACE WITH ValueStopwatch!
+                DependencyInjectionEventSource.Log.RealizedService(callSite, instance, timer);
```
pakrym: Instance would always be `Func<ServiceProviderEngineScope, object>` it's not very useful to trace it's type.

anurse:
Closing as per some in person discussions. Going to put this in the hands of @pakrym who is more aware of the issues we could actually be able to diagnose.

Issue #614, opened 2017-11-27
BuildServiceProvider should return IServiceProvider instead of ServiceProvider


cknaap:
Method Microsoft.Extensions.DependencyInjection.ServiceCollectionContainerBuilderExtensions.BuildServiceProvider returns a ServiceProvider instead of the associated interface IServiceProvider.

The latter allows for better mocking in unit tests.

Plus, the documentation on the method also says it returns *I*ServiceProvider.

davidfowl:
No, it shouldn't. How can you mock an extension method? This method is about building a concrete implementation of an IServiceProvider.

> Plus, the documentation on the method also says it returns IServiceProvider.

It was changed in 2.0 and needs to be updated /cc @Rick-Anderson

aspnet-hello:
This issue was moved to aspnet/Home#2329

PR #616, opened 2017-11-28
Update service provider type in API docs

henkmollema:
Fixes https://github.com/aspnet/Docs/issues/4872 and https://github.com/aspnet/DependencyInjection/issues/614


henkmollema:
Oops, too late 😸

PR #619, opened 2017-12-15
Add method to replace services with a new implementation type

### LINKED ISSUE #618: Replace services independent from their lifetime
Hi,

I was wondering if it possible to replace a service implementation independently from its service lifetime..? For now, afaik, one is only able to call the `IServiceCollection.Replace(...)` - extension providing the lifetime of the service to be replaced.

The point is, that I only would like to specify the lifetime of a service once and only once. Imagine the following:

```
public IServiceCollection UseExtraCostlyDependencies(this IServiceCollection services) {
	services.AddSingleton(...);
	services.AddSingleton(...);
	services.AddTransient(...);
	services.AddScoped(...);
}
```

Looking at other extension methods to use fake implementations for those costly dependencies, I'm forced to replace them dependent on their lifetime:

```
public IServiceCollection UseFakesForExtraCostlyDependencies(this IServiceCollection services) {
	services.Replace(ServiceDescriptor.Singleton(...));
	services.Replace(ServiceDescriptor.Singleton(...));
	services.Replace(ServiceDescriptor.Transient(...));
	services.Replace(ServiceDescriptor.Scoped(...));
}
```
Think of them as being serveral extensions methods, I'm just too lazy to write them all down - but I guess you get the point.

Long story short: The higher the count of real/fake-implementation - correlations, the higher the chance to mismatch a service lifetime when replacing them with fakes. This would potentially lead to inconsistet application behavior in different stages.

Is there maybe already a solution to my problem? Do I maybe apply the pattern wrong?

kind regards,
Peter

khellang:
I exposed `TryGetDescriptors` as well. It's a method that I've found useful a number of times. Let me know if you want to make it private instead.

Closes https://github.com/aspnet/Home/issues/2328

### DIFF
--- a/src/DI.Abstractions/Extensions/ServiceCollectionDescriptorExtensions.cs
+++ b/src/DI.Abstractions/Extensions/ServiceCollectionDescriptorExtensions.cs
@@ -685,6 +685,110 @@ namespace Microsoft.Extensions.DependencyInjection.Extensions
         }
 
         /// <summary>
+        /// Replaces all registrations for <typeparamref name="TService"/> with
+        /// new registrations using <typeparamref name="TImplementation"/>.
+        /// </summary>
+        /// <typeparam name="TService">The type of the services to replace.</typeparam>
+        /// <typeparam name="TImplementation">The new implementation type.</typeparam>
+        /// <param name="services">The <see cref="IServiceCollection"/>.</param>
+        public static IServiceCollection Replace<TService, TImplementation>(this IServiceCollection services)
+            where TImplementation : TService
+        {
+            return services.Replace<TService>(typeof(TImplementation));
+        }
+
+        /// <summary>
+        /// Replaces all registrations for <typeparamref name="TService"/> with
+        /// new registrations using <paramref name="implementationType"/>.
+        /// </summary>
+        /// <typeparam name="TService">The type of the services to replace.</typeparam>
+        /// <param name="services">The <see cref="IServiceCollection"/>.</param>
+        /// <param name="implementationType">The new implementation type.</param>
+        public static IServiceCollection Replace<TService>(this IServiceCollection services, Type implementationType)
+        {
+            return services.Replace(typeof(TService), implementationType);
+        }
+
+        /// <summary>
+        /// Replaces all registrations for <paramref name="serviceType"/> with
+        /// new registrations using <paramref name="implementationType"/>.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/>.</param>
+        /// <param name="serviceType">The type of the services to replace.</param>
+        /// <param name="implementationType">The new implementation type.</param>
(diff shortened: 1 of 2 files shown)


aspnet-hello:
This issue was moved to aspnet/Home#2327

aspnet-hello:
Sorry I was a bad bot. Re-opening this PR.

src/DI.Abstractions/Extensions/ServiceCollectionDescriptorExtensions.cs
```diff
@@ -684,6 +684,110 @@ public static void TryAddSingleton<TService>(this IServiceCollection collection,
...
+            return services.TryGetDescriptors(typeof(TService), out descriptors);
+        }
+
+        /// <summary>
+        /// Tries to get all registered descriptors with the specified <paramref name="serviceType"/>.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/>.</param>
+        /// <param name="serviceType">The type of the services to get.</param>
+        /// <param name="descriptors">The matching <see cref="ServiceDescriptor"/>s.</param>
+        public static bool TryGetDescriptors(this IServiceCollection services, Type serviceType, out ICollection<ServiceDescriptor> descriptors)
```
davidfowl: I'd make it internal.

src/DI.Abstractions/Extensions/ServiceCollectionDescriptorExtensions.cs
```diff
@@ -684,6 +684,110 @@ public static void TryAddSingleton<TService>(this IServiceCollection collection,
...
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            if (serviceType == null)
+            {
+                throw new ArgumentNullException(nameof(serviceType));
+            }
+
+            return (descriptors = services.Where(service => service.ServiceType == serviceType).ToList()).Count > 0;
```
davidfowl: Split this up into multiple lines. Treating assignment like an expression always looks messy IMO.

davidfowl:
- Where are the overloads that take a factory? Why only implementation type?
    ```C#
    services.AddSingleton<IFoo, Foo1>();

    services.Replace<IFoo>(sp => new Foo2());
    ```

- What's the behavior when you replace an implementation but there are many descriptors that exists for:
    ```C#
    services.AddSingleton<IFoo, Foo1>();
    services.AddSingleton<IFoo, Foo2>();

    services.Replace<IFoo>(sp => new Foo3());
    ```

khellang:
> Where are the overloads that take a factory? Why only implementation type?

Good point. I've only ever needed the overloads taking a type. I can add the factory overloads if you think they'd be valuable.

> What's the behavior when you replace an implementation but there are many descriptors that exists for:

In that case, you'd end up with two registrations of `IFoo`, both resolving `Foo3` as a singleton 🤔

muratg:
We're closing this PR because no updates have been provided in a timely manner and we are unable to merge it.

We thank you for the contribution and look forward to collaborating more in the future.

PR #624, opened 2018-01-27
Make TService type param more permissive on Add*<TService, TImplementation>()

Mardoxx:
Fixes aspnet/Home#2828

### DIFF
--- a/src/DI.Abstractions/ServiceCollectionServiceExtensions.cs
+++ b/src/DI.Abstractions/ServiceCollectionServiceExtensions.cs
@@ -87,7 +87,6 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <returns>A reference to this instance after the operation has completed.</returns>
         /// <seealso cref="ServiceLifetime.Transient"/>
         public static IServiceCollection AddTransient<TService, TImplementation>(this IServiceCollection services)
-            where TService : class
             where TImplementation : class, TService
         {
             if (services == null)
@@ -185,7 +184,6 @@ namespace Microsoft.Extensions.DependencyInjection
         public static IServiceCollection AddTransient<TService, TImplementation>(
             this IServiceCollection services,
             Func<IServiceProvider, TImplementation> implementationFactory)
-            where TService : class
             where TImplementation : class, TService
         {
             if (services == null)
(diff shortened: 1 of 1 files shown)


Mardoxx:
Will add tests after feedback (if this is a genuine issue)

davidfowl:
Is this a breaking change?

Mardoxx:
Not entirely sure. Superficially I don't think so!

Mardoxx:
Okay, it's been picked up as a breaking change - but is it really?

This is a strange one as it clearly works when [not passing template params](https://github.com/aspnet/DependencyInjection/blob/5238b8ccf7bc6592a5c6a1da8635ed2871cfabbf/src/DI.Specification.Tests/DependencyInjectionSpecificationTests.cs#L412).

What's different between

this
```cs
    public static IServiceCollection RegisterScopedTestService<TService, TImplementation>(this IServiceCollection services)
        where TImplementation : class, TService
    {
        return services.AddScoped<TService, TImplementation>();
    }
```

and
    public void SomeFunction(IServiceCollection services)
    {
        services.AddScoped<ITestService, TestService>();
    }

given
```cs
    public interface ITestService
    {
    }

    public class TestService : ITestService
    {
    }
```

The signature still expects TService to be a reference type, but the latter is working. What am I missing here?

davidfowl:
You just need to add `where TService : class` so that you match the constraints of the method you're calling.

Mardoxx:
Just realised that! 😄 

Now I'm wondering if the constraint is actually required. Under what circumstances could you register something which fails either?

```
TImplementation : class
TImplementation : TService, class
```

but passes with

```
TImplementation : class
TImplementation : TService, class
```

davidfowl:
> Now I'm wondering if the constraint is actually required. Under what circumstances could you register something which fails these constraints

It may not be possible but I'm not sure it matters because we wouldn't make a breaking change like this regardless.

Mardoxx:
interesting. I'll post on SO, see if anyone can out of curiosity.

Very understandable! Thanks 😄

Mardoxx:
So after a little reading it would seem that the current definition is more technically correct, even if it's not necessary. I suspect the choice was taken to have both constraints there as it clearly shows that TService must be a reference type!

Mardoxx:
https://stackoverflow.com/a/48471755/3515174 Brilliant insight here 😄

Issue #456, opened 2016-10-13
Revisit tracking transient services for disposal


khellang:
Tracking transient objects for disposal is a pretty bad practice and will almost always lead to memory leaks. Even if it's tracked in a scoped provider, it'll be kept around for longer than it needs to, at a minimum.

This is why containers like SimpleInjector and StructureMap (and probably others as well) _don't_ track transient disposables. SimpleInjector even has [a diagnostic warning](https://simpleinjector.readthedocs.io/en/latest/disposabletransientcomponent.html) about this.

In order to comply with the MS.Ext.DI specs, StructureMap had to [introduce a way to opt-in to track transients for disposal](http://structuremap.github.io/the-container/disposing/), which sucks.

I've already seen several reports on severe memory leaks because of this in the wild. Here's the latest example: https://gitter.im/aspnet/Home?at=57ff6e91891a530163074f94

> Hey guys ... anybody here that has some experience with the .NET Core DI implementation? I run into a memory leak because a type I registered with `services.AddTransient<IMyType, MyType>()` does not get garbage collected. With VS memory diagnostics I found out that a `List<IDisposable>` within the `ServiceProvider` class holds references to all instances of it. So it might be a misconception about how it should work on my side ... but I think that `ServiceProvider` should not hold references to types I registered as transient. Am I wrong about that?

I also suspect there are other, less severe memory leaks around, that people haven't noticed yet or worked around.

What can be done about this? The least breaking change I can see is add a `Release` method somewhere that can be used to explicitly evict specific instances from tracking, but ideally, this behavior should be removed altogether and throw if a transient disposable is registered.

// @davidfowl @pakrym @tillig @alexmg @nblumhardt @seesharper @jeremydmiller @dotnetjunkie

khellang:
Oh, I just noticed this is being "thought about" for a future version. :+1: (https://github.com/aspnet/DependencyInjection/issues/433#issuecomment-243360317) I'll keep this open for tracking this specific issue.

nblumhardt:
Watching this thread with interest - spent a long time on this topic over the years :-)

> Tracking transient objects for disposal is a pretty bad practice and will almost always lead to memory leaks.

This statement may not be the best starting point. It's one of those cases where "the way you've always done it" seems like the most logical way, whichever camp you're in. The alternative point of view goes something like - _how do you do release resources deterministically in a loosely-coupled system if your IoC container can't/won't do it?_

[...]

PR #625, opened 2018-01-29
Expanded captive dependency check

hvanbakel:
Captive dependency checking checked for singletons being dependent upon a scoped dependency. However, depending on a dependency which lifetime is shorter than its parent is not specific to that relation. This PR adds checks for Singleton -> Transient and Scoped -> Transient.

### DIFF
--- a/src/DI/Properties/Resources.Designer.cs
+++ b/src/DI/Properties/Resources.Designer.cs
@@ -150,6 +150,20 @@ namespace Microsoft.Extensions.DependencyInjection
         internal static string FormatDirectScopedResolvedFromRootException(object p0, object p1)
             => string.Format(CultureInfo.CurrentCulture, GetString("DirectScopedResolvedFromRootException"), p0, p1);
 
+        /// <summary>
+        /// Cannot consume {2} service '{0}' from {3} '{1}'.
+        /// </summary>
+        internal static string CaptiveDependencyException
+        {
+            get => GetString("CaptiveDependencyException");
+        }
+
+        /// <summary>
+        /// Cannot consume {2} service '{0}' from {3} '{1}'.
+        /// </summary>
+        internal static string FormatCaptiveDependencyException(object p0, object p1, object p2, object p3)
+            => string.Format(CultureInfo.CurrentCulture, GetString("CaptiveDependencyException"), p0, p1, p2, p3);
+
         private static string GetString(string name, params string[] formatterNames)
         {
             var value = _resourceManager.GetString(name);
--- a/src/DI/Resources.resx
+++ b/src/DI/Resources.resx
@@ -150,4 +150,7 @@
   <data name="DirectScopedResolvedFromRootException" xml:space="preserve">
     <value>Cannot resolve {1} service '{0}' from root provider.</value>
   </data>
+  <data name="CaptiveDependencyException" xml:space="preserve">
+    <value>Cannot consume {2} service '{0}' from {3} '{1}'.</value>
+  </data>
 </root>
--- a/src/DI/ServiceLookup/CallSiteValidator.cs
+++ b/src/DI/ServiceLookup/CallSiteValidator.cs
@@ -43,6 +43,25 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
 
         protected override Type VisitTransient(TransientCallSite transientCallSite, CallSiteValidatorState state)
         {
+            if (state.Singleton != null)
+            {
+                throw new InvalidOperationException(Resources.FormatCaptiveDependencyException(
+                    transientCallSite.ServiceType,
+                    state.Singleton.ServiceType,
+                    nameof(ServiceLifetime.Transient).ToLowerInvariant(),
+                    nameof(ServiceLifetime.Singleton).ToLowerInvariant()
+                    ));
+            }
+
+            if (state.Scoped != null)
+            {
+                throw new InvalidOperationException(Resources.FormatCaptiveDependencyException(
+                    transientCallSite.ServiceType,
+                    state.Singleton.ServiceType,
+                    nameof(ServiceLifetime.Transient).ToLowerInvariant(),
+                    nameof(ServiceLifetime.Scoped).ToLowerInvariant()
+                    ));
+            }
             return VisitCallSite(transientCallSite.ServiceCallSite, state);
         }
 
@@ -88,6 +107,8 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
             {
                 return null;
             }
+
+            state.Scoped = scopedCallSite;
             if (state.Singleton != null)
             {
                 throw new InvalidOperationException(Resources.FormatScopedInSingletonException(
@@ -113,6 +134,8 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
         internal struct CallSiteValidatorState
         {
             public SingletonCallSite Singleton { get; set; }
+
+            public ScopedCallSite Scoped { get; set; }
         }
     }
 }
--- a/test/DI.Tests/ServiceProviderValidationTests.cs
+++ b/test/DI.Tests/ServiceProviderValidationTests.cs
@@ -23,6 +23,35 @@ namespace Microsoft.Extensions.DependencyInjection.Tests
         }
 
         [Fact]
+        public void GetService_Throws_WhenTransientIsInjectedIntoSingleton()
+        {
+            // Arrange
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddSingleton<IFoo, Foo>();
+            serviceCollection.AddTransient<IBar, Bar>();
+            var serviceProvider = serviceCollection.BuildServiceProvider(validateScopes: true);
+
+            // Act + Assert
+            var exception = Assert.Throws<InvalidOperationException>(() => serviceProvider.GetService(typeof(IFoo)));
+            Assert.Equal($"Cannot consume transient service '{typeof(IBar)}' from singleton '{typeof(IFoo)}'.", exception.Message);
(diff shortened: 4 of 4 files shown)


hvanbakel:
So this expands the check from just Scoped dependencies to a lifetime check.

khellang:
Even though this highlights potential bugs/leaks in your code, and the scope checking isn't enabled by default outside the `Development` environment, this is a breaking change. I wonder why @pakrym didn't add these checks in the first place?

hvanbakel:
@khellang correct, so maybe adding this as a separate option is a better idea as that would make it non-breaking

khellang:
Looks like it might've just been overlooked/forgotten... https://github.com/aspnet/DependencyInjection/pull/430

khellang:
I guess it could be kept under the same flag, but using something similar to MVC's [CompatibilitySwitch](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Infrastructure/CompatibilitySwitch.cs). Adding flags to methods gets really messy quickly.

pakrym:
Capturing transitive from singletons/scopes is often completely valid scenario, this would cause too many false positives.

hvanbakel:
@pakrym how is that the case? That is exactly the behavior that you're trying to prevent by setting the ValidateScopes flag to true?

A transitive in a singleton becomes a singleton.

pakrym:
Often transitive is not more then "I want a new instance every time because I have internal state and don't care about how long my object would be alive". Also, this change would make transitive useless in ASP.NET Core, because you are always in request scope.

hvanbakel:
So if the abstraction is that you get a new instance every single time, how does that hold if you're resolving a transient dependency inside a singleton?

Are you saying that every dependency in my controller has request scope?

hvanbakel:
Ok so let's assume that as the implementor of an interface, I do not know who takes a dependency on me (as it should) and I would like to use internal state. How do I guarantee that one of my consumers does not make me a singleton scoped instance?

Take this below code for example, when I'm developing ITransientDependency I am making the assumption that I will be instantiated every single time. Isn't this exactly what the developer of for example IOptionsSnapshot<T> assumes?

```
public class TestController : Controller
    {
        private readonly ISingletonDependency dependency;

        public TestController(
            ISingletonDependency dependency)
        {
            this.dependency = dependency;
        }

        [HttpGet]
        [Route("")]
        public int AreEqual()
        {
            return this.dependency.Dependency.Count;
        }
    }

    public interface ISingletonDependency
    {
        ITransientDependency Dependency { get; }
    }

    public class SingletonDependency : ISingletonDependency
    {
        public SingletonDependency(ITransientDependency dependency)
        {
            this.Dependency = dependency;
        }

        public ITransientDependency Dependency { get; private set; }
    }

    public interface ITransientDependency
    {
        int Count { get; }
    }

    public class TransientDependency : ITransientDependency
    {
        private int count = 0;

        public int Count => ++this.count;
    }
```

pakrym:
> How do I guarantee that one of my consumers does not make me a singleton scoped instance?

Why would you care about this?

hvanbakel:
Let me turn the question around, why do you care about a singleton taking a dependency on a scoped dependency?
You care because it breaks what you were trying to do. Take a look at `IOptionsSnapshot<T>` for example, that is an example of where the current check throws as it is trying to read on every request and wrapping that in a singleton makes it break.

pakrym:
Because we've seen bugs caused by it and it wasn't aggressively breaking a lot of code. Also transient does not denote how "long" the services would live just that it "Transient lifetime services are created each time they're requested. "

There was a discussion about capturing IDisposible instances in root container: https://github.com/aspnet/DependencyInjection/issues/456

Did you try running MVC app with this checks enabled?

hvanbakel:
I did, it doesn't fail and it increases the count with every request. Here's my program.cs:
```
    public class Program
    {
        public static void Main(string[] args)
        {
            BuildWebHost(args).Run();
        }

        public static IWebHost BuildWebHost(string[] args) =>
            WebHost.CreateDefaultBuilder(args)
                .UseDefaultServiceProvider((context, options) =>
                {
                    options.ValidateScopes = context.HostingEnvironment.IsDevelopment();
                })
                .UseStartup<Startup>()
                .Build();
    }
```
and startup

```
    public class Startup
    {
        // This method gets called by the runtime. Use this method to add services to the container.
        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddMvc();

            services.AddTransient<ITransientDependency, TransientDependency>();
            services.AddScoped<IScopedDependency, ScopedDependency>();
            services.AddSingleton<ISingletonDependency, SingletonDependency>();
        }

        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }

            app.UseMvc();
        }
    }
```

pakrym:
Strange. There are a lot of MVC singleton services that cache transients:

https://github.com/aspnet/Mvc/blob/1d6b02c1f56ef46bc4c8568e748747c82bbf1bb8/src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs#L187-L190

hvanbakel:
So the issue here is that nesting a transient in a singleton does not fail.

So singleton > scoped > transient (in lifetime)
So nesting a scoped in a singleton expands the lifetime of the scoped dependency and crashes
Nesting a transient in a singleton also expands the lifetime but doesn't crash
The line above is what this PR addresses

pakrym:
@hvanbakel  Did you try running MVC with your additional check enabled?

hvanbakel:
I did, so now I get:
```
+		$exception	{System.InvalidOperationException: Cannot consume transient service 'Microsoft.Extensions.Options.IOptionsFactory`1[Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions]' from singleton 'Microsoft.Extensions.Options.IOptions`1[Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions]'.
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.VisitTransient(TransientCallSite transientCallSite, CallSiteValidatorState state)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSite(IServiceCallSite callSite, TArgument argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.VisitConstructor(ConstructorCallSite constructorCallSite, CallSiteValidatorState state)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSite(IServiceCallSite callSite, TArgument argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.VisitSingleton(SingletonCallSite singletonCallSite, CallSiteValidatorState state)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSite(IServiceCallSite callSite, TArgument argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.VisitConstructor(ConstructorCallSite constructorCallSite, CallSiteValidatorState state)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSite(IServiceCallSite callSite, TArgument argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.VisitSingleton(SingletonCallSite singletonCallSite, CallSiteValidatorState state)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSite(IServiceCallSite callSite, TArgument argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.ValidateCallSite(IServiceCallSite callSite)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.Microsoft.Extensions.DependencyInjection.ServiceLookup.IServiceProviderEngineCallback.OnCreate(IServiceCallSite callSite)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngine.CreateServiceAccessor(Type serviceType)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngine.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngine.GetService(Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at Microsoft.AspNetCore.Hosting.Internal.WebHost.EnsureServer()
   at Microsoft.AspNetCore.Hosting.Internal.WebHost.BuildApplication()}	System.InvalidOperationException
```

which is good. Because this signals that `Microsoft.Extensions.Options.IOptionsFactory` now has singleton lifetime inside of `Microsoft.Extensions.Options.IOptions`. 

It's a bit odd to crash for singleton wrapping a scoped but not for this though as it is essentially the same problem of 1 component with a long lifetime wrapping a component with a short lifetime.

PR #626, opened 2018-01-31
Prevent resolutions after calling Dispose

pakrym:


test/DI.Tests/ServiceProviderContainerTests.cs
```diff
@@ -116,13 +116,65 @@ public void DoesNotDisposeSingletonInstances()
...
+        {
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddScoped<IFakeService, FakeService>();
+
+            var provider = CreateServiceProvider(serviceCollection);
+            var scope = provider.CreateScope();
+            scope.Dispose();
+
+            Assert.Throws<ObjectDisposedException>(() => scope.ServiceProvider.GetService<IFakeService>());
+            //Chec that resolution from root works
```
davidfowl: Check

src/DI/ServiceLookup/ServiceProviderEngine.cs
```diff
@@ -57,16 +86,11 @@ protected ServiceProviderEngine(IEnumerable<ServiceDescriptor> serviceDescriptor
             return _ => null;
         }
 
-        internal object GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
halter73: Did you compare perf inlined vs non-inlined?

src/DI/ServiceLookup/ServiceProviderEngineScope.cs
```diff
@@ -26,6 +26,11 @@ public ServiceProviderEngineScope(ServiceProviderEngine engine)
 
         public object GetService(Type serviceType)
         {
+            if (_disposed)
+            {
+                throw new ObjectDisposedException(nameof(IServiceScope));
```
halter73: Why change patterns from ServiceProviderEngine and throw ODE directly from the method?

test/DI.Tests/ServiceProviderContainerTests.cs
```diff
@@ -116,13 +116,65 @@ public void DoesNotDisposeSingletonInstances()
...
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddScoped<IFakeService, FakeService>();
+
+            var provider = CreateServiceProvider(serviceCollection);
+            var scope = provider.CreateScope();
+            scope.Dispose();
+
+            Assert.Throws<ObjectDisposedException>(() => scope.ServiceProvider.GetService<IFakeService>());
+            //Check that resolution from root works
+            Assert.NotNull(provider.CreateScope());
```
halter73: Nit: Comment doesn't seem to match assertion since you aren't resolving anything from the root.

halter73 (approved):

test/DI.Tests/ServiceProviderContainerTests.cs
pakrym: Create scope resolves scope factory

src/DI/ServiceLookup/ServiceProviderEngineScope.cs
pakrym: Good point.

src/DI/ServiceLookup/ServiceProviderEngine.cs
pakrym: No, but we usually extract throwing into non inlined methods so their callers are not increasing in size and inline easier.

davidfowl (approved):

test/DI.Tests/ServiceProviderContainerTests.cs
halter73: 👍

PR #632, opened 2018-03-19
Remove extra references

pakrym:


natemcmaster (approved):

ryanbrandenburg (commented):
Doesn't this require you to remove the rest of the changes in https://github.com/aspnet/DependencyInjection/commit/d5b3851ea5f53c6acd7437b91251da8db7c63076 too? If not why did we add these refs in the first place.

pakrym:
These refs are included in net461 and netcoreapp2.0/2.1 that DI is cross-compiled to.

PR #636, opened 2018-04-13
Disable IL backend by default

pakrym:
https://github.com/aspnet/Home/issues/3054 seems to be severe and can affect existing customers.


Eilon:
So we're keeping the feature in general? E.g. no need to remove https://github.com/aspnet/DependencyInjection/blob/2d3decdfe47b7e685c41b2bf082ac138538c2cc0/src/DI/ServiceProvider.cs#L35 ?

davidfowl (approved):

davidfowl:
> So we're keeping the feature in general? E.g. no need to remove

It's not being removed completely, we're just removing the default for the "compiled" mode. Before it would default to ref emit based on the platform.

Eilon:
OK

PR #637, opened 2018-04-25
Unit tests for aspnet/Home#2871 and aspnet/Home#3091

fubar-coder:
This unit tests show, that the problem described by aspnet/Home#2871 was only partially fixed.

The unit tests fail, but that was expected to show the problem described in aspnet/Home#2871 and aspnet/Home#3091


muratg:
@pakrym could you take a quick look when you have some time please?

pakrym:
Closing because of the reasons stated in https://github.com/aspnet/AspNetCore/issues/3091

PR #638, opened 2018-04-25
Make lifetime a callsite property and add stackguards

pakrym:
Changing cache key from `Tuple<CallSite, Type>` allows us:
1. To create callsites non-atomically
2. Remove the global lock in call site factory
3. Add a stack guard that dispatched instead of throwing SO
4. Emit key directly in IL instead of passing reference through

The cache key is changed to `(type, slot)` pair where `type` is service type and `slot` is an index of the item in case it was resolved in `IEnumerable<Type>`

Additional changes:
1. Removed create instance call site - it's just a subset of ConstructorCallSite

### DIFF
--- a/src/DI/DI.csproj
+++ b/src/DI/DI.csproj
@@ -10,7 +10,7 @@
     <DefineConstants Condition="'$(ILEmitBackend)' == 'True'">$(DefineConstants);IL_EMIT</DefineConstants>
 
     <!-- Debug IL generation -->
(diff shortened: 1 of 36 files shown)


davidfowl (approved):

muratg:
Any impact on user code (i.e. is it a breaking change for users?)

pakrym:
No

muratg:
@pakrym could you either merge or close this? @natemcmaster is going to move this to the mondo repo and moving PRs are tricky. Thanks.

PR #661, opened 2018-10-22
Dispose service instances only once

poke:
This is a fix for aspnet/AspNetCore#2932. Unfortunately, just changing the collection to a hash set would not work as #505 requires this to be an ordered collection. So instead, I’m remembering what services I have disposed at the time they get disposed.

An alternative solution would be to have the hash set around as an instance member to avoid *adding* duplicate services to the collection. This would have the benefit that services are disposed in the reversed order of their *first* use. With this fix, they are disposed in the order of their *last* use. Not sure which of those is better; this version at least avoids having a second collection around all the time.


natemcmaster:
@poke thanks for the PR. I'm going to close this because we're merging this repo with https://github.com/aspnet/Extensions. After https://github.com/aspnet/DependencyInjection/pull/662 has been merged, please feel free to re-open this PR over there.

cc @davidfowl

PR #663, opened 2018-11-02
Do not throw circular reference exception when service resolved as IEnumerable and not in one callsite

pakrym:
https://github.com/aspnet/AspNetCore/issues/3232

Easier to review with whitespace ignored.


natemcmaster:
Sorry, for bad timing. I just merged this repo with aspnet/Extensions. Closing as this should be reopened on the new home for DI.

